PR #28, opened 2022-12-27
add customuser, accounts app

RachellCalhoun:


RachellCalhoun:
I left the calendar models out because we probably want to look into various options. 
Speakers -- I can see as being one time guests and not necessarily wanting to create a whole profile on our platform. So I left them as nonusers but we can change this later if we need.

accounts/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class CustomUser(AbstractUser):
+    PARTICIPANT = "Participant"
+    MENTOR = "Mentor"
+    EXPERT = "Expert"
+    PROJECT_OWNER = "Project Owner"
+    VOLUNTEER = "Volunteer"
+    ORGANIZER = "Organizer"
+    MEMBER_ROLES = (
```
dawnwages, line 16: nice!

accounts/admin.py
```diff
@@ -0,0 +1,54 @@
+import csv
+
+from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
+from django.http import HttpResponse
+
+from accounts.models import Link, CustomUser
+
+
```
dawnwages, line 9: dope :) I like it!

indymeet/templates/wagtailusers/users/create.html
```diff
@@ -0,0 +1,11 @@
+{% extends "wagtailusers/users/create.html" %}
```
dawnwages, line 1: How do I get to this form? Is it part of `wagtail_urls`? I want to link it with the front end buttons I created for it already.

dawnwages:
Can merge, let me know how to get to the forms you created, and I can make the button do the thing.

indymeet/templates/wagtailusers/users/create.html
RachellCalhoun: Yes, I think it's part of wagtail_urls, I followed this documentation https://docs.wagtail.org/en/stable/advanced_topics/customisation/custom_user_models.html

PR #32, opened 2022-12-29
22 change db

dawnwages:


requirements/requirements.txt
```diff
@@ -1,2 +1,4 @@
 Django>=4.1,<4.2
 wagtail>=4.1,<4.2
+django-storages
+psycopg2-binary
```
RachellCalhoun, line 4: Did you mean to remove django-bootstrap? I will merge as is, but please check and update if required.

dawnwages: I need django-bootstrap too. I don't know why I forgot this. `django-storages` is needed for Azure deploy.

PR #53, opened 2023-01-07
add recaptcha, keys

### LINKED ISSUE #34: Verify users on registration

RachellCalhoun:
closes #34 adding recaptcha to the user registration form. 

Since I'm adding keys here, I created a .env.example , this file is used to show the environmental variables that we store in .env file. 

To faciliate this and to continue with our `os.getenv` method, I've used python-dotenv, there are others : django-environ, django-dotenv etc, and we can change at a later date if required, but I needed something in this PR.


dawnwages (approved):

dawnwages:
I'll add the values to azure and when we set up the CI/CD and merge develop into main, it'll already be there. Is there a problem when developing locally without it?

RachellCalhoun:
@dawnwages yeah good catch, let me make sure it's going to work without recaptcha locally.

.env.example
```diff
@@ -0,0 +1,2 @@
+RECAPTCHA_PUBLIC_KEY=dummy_value
+RECAPTCHA_PRIVATE_KEY=dummy_value
```
RachellCalhoun, line 2: @dawnwages  :point_up:

PR #57, opened 2023-01-16
22 change db

dawnwages:
Change db and update settings for production.

- Changed collect files output directory name from `static` => `staticfiles` so you'll need to run `collectstatic` and delete the `static` folder locally.


indymeet/settings/dev.py
```diff
@@ -11,6 +11,13 @@
         }
     }
 
+print('----------------------------------')
+print('----------------------------------')
+print('DEV')
+print('----------------------------------')
+print('----------------------------------')
+
+
```
RachellCalhoun, line 20: @dawnwages Was this intentionally left in?

RachellCalhoun (approved):
I left some comments for what appear to be some forgotten about print statements but other than that looks good to me.

indymeet/settings/dev.py
dawnwages: Yea it was but I can take them out if you think they're ugly.

RachellCalhoun: Nope, you are good. Keep it in, i just thought you might have forgotten. :+1:

PR #77, opened 2023-01-24
add events in template, add tags

### LINKED ISSUE #71: Populate Event cards with real events from db

### LINKED ISSUE #76: Change event categories to tags

RachellCalhoun:
- changed categories to wagtail tags 
- populated card templates 
- created list view for all events with tag filtering 
- created queryset managers for events
- rsvp/cancel rsvp from homepage if logged in and coc accepted and event is in future

closes #76 
closes #71


RachellCalhoun:
video walk through:  https://www.loom.com/share/e31932f5bde04223b28d83eda89d67b9

Added to profile as well: 
image

RachellCalhoun:
Also, I wasn't sure the best way forward with how we are creating events. Right now I'm just using the django admin. It might be nice to use wagtail? But since I'm not as familiar and the m2m will need more work on wagtail admin, where as it doesn't on django admin i left i as is. This might be something to think about before we populate the db with test data. THanks!

home/managers.py
```diff
@@ -0,0 +1,28 @@
...
+        return self.filter(status=self.model.PENDING)
+
+    def visible(self):
+        return self.exclude(status=self.model.PENDING)
+
+    def upcoming(self):
+        return self.filter(start_time__gte=timezone.now())
+
+    def past(self):
+        print('time zone now', timezone.now())
```
sabderemane, line 26: maybe to remove ?

home/templates/home/includes/event_card.html
```diff
@@ -6,22 +6,40 @@
...
-          {% include 'home/includes/event_categories.html' %}   
-          <p class="card-text pt-4">Some quick example text to build on the card title and make up the bulk of the card's content.</p>
+          {% include 'home/includes/event_categories.html' %}
+          <p class="card-text pt-4">{{ event.description }}</p>
         {% include 'home/includes/event_team.html' %}
         </div>
         <div class="card-footer">
-            <a href="#" class="btn btn-primary">About</a>
+            <a href="{% url 'event_detail' pk=event.id %}" class="btn btn-primary">About</a>
             {% if cat == "upcoming" %}
```
sabderemane, line 35: it shouldn't be `tag` (or something like that) instead of `cat` ?

RachellCalhoun: Good call out, I decided to remove it since we have a manager to determine past/upcoming.

sabderemane (approved):
LGTM, may be Dawn will see something else

dawnwages:
This looks great! sorry it took so long for review.

dawnwages:
I'll fix the merge conflicts

PR #130, opened 2023-11-15
Closing applications tweaks

sarahboyce:
I think as we want to have people sign up via the website in future, we can stop advertising it there for now. 
Small bug that we closed the applications on our website a day early, should fix this too. 🐛 

Don't worry about the event images 👍


RachellCalhoun (dismissed):
lgtm :rocket:

PR #167, opened 2023-12-27
add richtextblock

### LINKED ISSUE #166: Add TextBlock
I think we don't have a plain `TextBlock`, this would be useful for the blog. Also would be good for `TextBlock` to be `RichTextBlock` so we have more flexibility with editing. If there is better way, I'm open to it.

RachellCalhoun:
closes #166


tim-schilling (approved):
Looks good. Let the previous build finish for staging to confirm I didn't break something with my settings changes before merging this one.

PR #172, opened 2023-12-29
Update the text-only blocks.

tim-schilling:
The paragraph block was limited to a set number of characters. It makes more sense to allow a text field instead since the editor didn't limit it.

This also exposes the TextWithHeading block for text_with_heading


dawnwages:
ga! I didn't see this PR and I've solved this already. Check out the changes in #173 PR? Its all on develop branch and tell me how it looks. https://staging-djangonaut-space.azurewebsites.net/comms/elements

PR #175, opened 2023-12-30
Document how to create/load fixture data

tim-schilling:
This documents how to create a fixture for the application. This will make it easier for folks to share reproducible environments locally and avoid needing to rely on staging so heavily.

@dawnwages this also adds the database environment variable to the settings which should restore your local environment.

@sabderemane @RachellCalhoun @dawnwages here's a fixture to use that has the first draft of Rachell's blog post that's on production. Or at least some version of it.

[fixture-20231230.zip](https://github.com/djangonaut-space/wagtail-indymeet/files/13798670/fixture-20231230.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ node_modules
 build
 theme/static/css/dist/style.css
 
+# Fixtures until we know what we're doing
+fixtures/data.json
+
 # misc
 .DS_Store
 
--- a/README.md
+++ b/README.md
@@ -151,8 +151,8 @@ This is an example of how to list things you need to use the software and how to
 6. Run migrations and create superuser
    ```sh
    python manage.py migrate
-   ```
-   ```sh
+   # Potentially load data first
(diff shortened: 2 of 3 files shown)


RachellCalhoun (approved):
LGTM, I had a few comments below. Thanks! 

1. It might be helpful to add/change instructions to include loading fixtures to an existing db, which is what I did by skipping a few steps. I think this is also a very likely situation because if someone is setting up the project for the first time they've likely done the readme install which includes creating a db, or it's someone who has been working on in and has a db. I know this won't work for all cases. 
2.  should we add `fixtures/data.json` to .gitignore?

tim-schilling:
I'm skeptical about adding this to an existing db because it uses primary keys rather than natural keys. If there are conflicts with existing data, it could result in data being associated with the wrong objects.

Yeah, we probably should ignore it until we know we want to store something.

tim-schilling:
Amended my commit with the gitignore addition.

PR #176, opened 2024-01-01
Add blog link to the pages.

tim-schilling:
This means we're going to need a comms page in wagtail.


tim-schilling:
Oh, that's neat!

PR #177, opened 2024-01-02
This should allow tailwind to be built.

tim-schilling:
This stops using collectstatic as a part of the oryx build system and uses a custom post build script.

Since tailwind requires node to compile the final css file, we need to add a package.json file and ENABLE_MULTIPLEPLATFORM_BUILD to get Oryx to install both python and node. This allows the post build script to run.

This also fixes the Azure media/static file storage issue by exposing the environment variables to actually push content there.

This will need a few new secrets configured and we may potentially need to change AZURE_CUSTOM_DOMAIN since I don't think that was ever used/tested.

### DIFF
--- a/indymeet/settings/production.py
+++ b/indymeet/settings/production.py
@@ -40,8 +40,7 @@ if os.getenv("ENVIRONMENT") == "production":
     }
     # Azure Media and Static Storage Settings
     AZURE_ACCOUNT_NAME = os.environ.get("AZURE_ACCOUNT_NAME", "djangonaut")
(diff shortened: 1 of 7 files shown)


indymeet/settings/production.py
dawnwages, line 1: @tim-schilling Does switching this work? I did a bad job naming these but ran into a bug following other instructions and needed to swap these key names.

dawnwages (approved):

indymeet/settings/production.py
tim-schilling: Did we answer this on our call?

tim-schilling: I can't remember when we ended that versus when this was asked.

PR #193, opened 2024-01-11
Specify use_json_field for StreamField to silence warning.

tim-schilling:
I suspect we actually want to be using True, but the model defaulted to False and a text column behind the scenes.

If we're not using the GeneralPage in production or staging yet, should we use `use_json_field=True` instead? I can't imagine the conversion process will be easier down the road. @dawnwages @RachellCalhoun


tim-schilling:
I talked this over with Rachell. We don't have any general pages on production and the StreamField requires this in Wagtail 5, so it's best to get it out of the way now.

PR #203, opened 2024-01-14
Use pip-tools and dependabot

### LINKED ISSUE #202: Switch to pip-tools for dependency management
We should try to have a more robust way to update dependencies. I think we can start with `pip-tools` and if that's not working for us, we can move to something like `poerty` or `pipenv`. I like `pip-tools` because it's relatively simple and the upgrade command is included in the generated requirements.txt file.

tim-schilling:
Closes #202

pip-tools:
This avoids having underlying dependencies change between deployments.
This makes our infrastructure more robust and reliable.

Dependabot:
This will cover python, npm and github actions.

The python and npm dependencies will be grouped together. We have a number
of dependencies already, so I'd like to avoid flooding our PRs with dependency
updates. The GitHub actions are fairly small, so I suspect we can have those
broken out.


RachellCalhoun (approved):
lgtm! 

It might be useful to add some documentation in the readme for updating these.

RachellCalhoun (approved):

PR #205, opened 2024-01-14
Pin node version in package.json

tim-schilling:
This prevents a person from running manage.py tailwind install and updating the dependencies for a different version of node than Azure can support.

### DIFF
--- a/theme/static_src/package-lock.json
+++ b/theme/static_src/package-lock.json
@@ -1,8 +1,1530 @@
 {
   "name": "theme",
   "version": "3.6.0",
-  "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


sarahboyce:
https://github.com/djangonaut-space/wagtail-indymeet/issues/226
Can we upgrade nodejs instead? :)

tim-schilling:
Go for it!

sabderemane (approved):
I feel like this have been in another PR no?

tim-schilling:
I'm moving forward with this as a pre-cursor to attempting to get our build to be compatible with a recent version of node.

PR #218, opened 2024-01-15
Add basic playwright test

tim-schilling:
This adds two tests:
- Smoke test of web app
- Smoke test of wagtail application

This also adds a local.sh helper script to run runserver and tailwind start in one file. It's not usable by windows folks, but it should provide a guide to implementing one for windows.

The testing auth flow is to have a predefined username and password for the playwright testing user. This means if you want to run the smoke tests against production locally, you'll need the password for the production playwright user. This is a bit risky considering the test user has Editor permissions to access the wagtail admin.

### DIFF
--- a/.env.template.local
+++ b/.env.template.local
@@ -5,3 +5,5 @@ RECAPTCHA_PRIVATE_KEY='dummy_value'
 # Only set to a non-empty string value if you want it enabled
 # The toolbar should not be used in a remote environment
 ENABLE_TOOLBAR=""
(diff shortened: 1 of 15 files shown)


RachellCalhoun (approved):
lgtm! 

Could you add a line in the readme or replace the shell commands to get the project started with the new local script?

tim-schilling:
@RachellCalhoun I can't replace them because I don't have a windows equivalent. I did add a mention of it though.

PR #219, opened 2024-01-15
Add actions to run smoke tests on staging and production.

tim-schilling:
Production is configured to only run on request since it does delete a blog and I'm not sure if we have backups yet.

### DIFF
--- a/.github/workflows/develop_end_to_end_tests.yml
+++ b/.github/workflows/develop_end_to_end_tests.yml
@@ -1,0 +1,44 @@
+name: Staging smoke test
+
+# Only trigger, when the build workflow succeeded
+on:
+  workflow_run:
+    workflows: ["Staging build and deploy Python app to Azure Web App - djangonaut-space"]
+    types:
+      - completed
+  workflow_dispatch:
+
+jobs:
+  build:
+    env:
+      ENVIRONMENT: 'test'
+      DJANGO_SETTINGS_MODULE: 'indymeet.settings.test'
+      DEBUG: False
+      # We don't need the app's secret key as long as we are only running our
+      # playwright tests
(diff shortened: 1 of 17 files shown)


.github/workflows/main_end_to_end_tests.yml
```diff
@@ -0,0 +1,40 @@
+name: Staging smoke test
+
+# Only trigger, when the build workflow succeeded
```
sabderemane, line 3:
```suggestion
# Manual trigger, when the build workflow succeeded
```
This is manual only for main or did I miss something?

.github/workflows/main_end_to_end_tests.yml
```diff
@@ -0,0 +1,40 @@
+name: Staging smoke test
```
sabderemane, line 1:
```suggestion
name: Main smoke tests
```

.github/workflows/develop_end_to_end_tests.yml
```diff
@@ -0,0 +1,44 @@
+name: Staging smoke test
```
sabderemane, line 1:
```suggestion
name: Staging smoke tests
```

sabderemane (commented):
Nice, just few wording comments

.github/workflows/main_end_to_end_tests.yml
tim-schilling: You're 100% right. I copy/pasted, adjusted and forgot to clean up.

tim-schilling:
Before this can be merged, we need to create the test user on the staging environment to confirm this will work.

sabderemane (approved):

tim-schilling:
This doesn't seem to be a priority any more

PR #267, opened 2024-02-14
Fixed accounts signup form.

sarahboyce:
Currently internal server error: https://djangonaut.space/accounts/signup/


tim-schilling:
@sarahboyce I confirmed that we're still running python 3.9 on the production server. The dependencies need to match that.

PR #269, opened 2024-02-15
Add bool for program update emails and update styling.

sarahboyce:
image

image


accounts/forms.py
```diff
@@ -19,13 +19,20 @@ class CustomUserCreationForm(UserCreationForm):
     receive_newsletter = forms.BooleanField(
```
sarahboyce, line 19: We could even remove this from the form actually? Unless we plan on doing a newsletter. 🤔

PR #280, opened 2024-02-17
Remove migrate from postbuild script.

tim-schilling:
This would require us to have the database connection settings in our github repository and that doesn't sit well with me currently. Let's keep looking.


tim-schilling:
Nope, it seems like the post build script is what we're supposed to use. https://github.com/Microsoft/Oryx/wiki/Django-Tips That's sort of frustrating.

tim-schilling:
This will be re-added in #282

PR #282, opened 2024-02-17
Switch to using DATABASE_URL for db connections.

### LINKED ISSUE #281: Switch to DATABASE_URL environment variable
We should use a database url value for our database connection information. It allows us to put more of our configuration options into a single environment variable. This makes it easier to connect to the database from other areas. Specifically configuring github to run migrations after builds for both staging and production (2 variables vs 8)

tim-schilling:
This reduces the number of environment variables we need for database connections. This will make environment configuration easier.

Closes #281

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,7 +1,5 @@
-DATABASE='<DATABASENAME>'
-PASSWORD='<PASSWORD>'
-USER='<USER>@<RESOURCE>'
-HOST='<HOST>.postgres.database.azure.com'
+#DATABASE_URL://USER:PASSWORD@HOST:PORT/NAME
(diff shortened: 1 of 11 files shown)


tim-schilling:
I've already added the new DATABASE_URL configuration to our github settings and azure app services for both production and staging. Once this is rolled out entirely, we can remove `USER`, `PASSWORD` and `HOST` from all of the configurations.

@dawnwages Do you remember what the `*_ADDON` settings were for in the github repo secrets?

tim-schilling:
From meeting with Dawn, the ADDON setting was from an additional container just for staging on how databases are managed.

PR #288, opened 2024-02-17
Install and use django-mailer

tim-schilling:
This will send emails out on a queue process using supervisord and postgres' listen/notify. Emails should go out every 30s by default.

### DIFF
--- a/indymeet/settings/base.py
+++ b/indymeet/settings/base.py
@@ -33,6 +33,7 @@ INSTALLED_APPS = [
     "accounts",
     "home",
     "search",
+    "mailer",
     "anymail",
     "captcha",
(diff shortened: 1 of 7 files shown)


sarahboyce (approved):
Nice! 🚀

tim-schilling:
This was a bust because it was incompatible with django-anymail (it was injecting the Message-ID header in a way that mailjet didn't like) and it was discovered that it uses pickle under the hood to load the message to be sent. Since there's an alternative, it doesn't make sense for us to absorb that type of security risk on the project. This was rolled back in 11b98674984d6b2b90234e8f47f52765aa9de5ce

PR #292, opened 2024-02-18
Updates based off PR review.

sarahboyce:


indymeet/templates/registration/login.html
```diff
@@ -23,13 +23,7 @@
...
             </div>
         {% endif %}
-        {% if form.errors and not form.non_field_errors %}
-          <div class="alert alert-danger">
-          {% for key, error in form.errors.items %}
-          <p> {{ key }}:{{ error }}</p>
-          {% endfor %}
-          </div>
-          {% endif %}
+        {% include 'includes/no_field_errors.html' %}
```
tim-schilling, line 26: Wait, isn't this only rendering the `non_field_errors` now though?

indymeet/settings/production.py
```diff
@@ -55,10 +55,7 @@
     STATIC_URL = f"https://{AZURE_CUSTOM_DOMAIN}/{AZURE_STATIC_CONTAINER}/"
     MEDIA_URL = f"https://{AZURE_CUSTOM_DOMAIN}/{AZURE_MEDIA_CONTAINER}/"
 
-    SENTRY_DNS = os.environ.get(
-        "SENTRY_DNS",
-        "https://302dde9a33fabc5af106ccf19ae23680@o4506747118157824.ingest.sentry.io/4506747129626624"  # Staging
-    )
+    SENTRY_DNS = os.environ.get("SENTRY_DNS")
```
tim-schilling, line 58: I created a new DSN for staging and added it as an environment variable to our azure staging environment. When you merge this, I'll delete the DSN we had previously.

indymeet/templates/registration/login.html
sarahboyce: Yes but the other errors are rendered under the fields, I don't think we need to put them twice.

tim-schilling: Ah, that makes sense. Good catch

tim-schilling (approved):
:rocket:

PR #295, opened 2024-02-19
Production deployment

sarahboyce:
- Added navigation to login/logout etc
- Added profile when logged in
- Updated form styling

See: https://staging-djangonaut-space.azurewebsites.net


tim-schilling (approved):
I gave the login, logout, profile views all a test. Looks good to me.

I do think we'd be better served by redirecting to the referrer for the login view, but it's not a major deal.

sarahboyce:
> I gave the login, logout, profile views all a test. Looks good to me.
> 
> I do think we'd be better served by redirecting to the referrer for the login view, but it's not a major deal.

I agree, I'll create an issue for that in case a Djangonaut wants to pick it up 👍

PR #300, opened 2024-02-19
Add daily management command to send notification email when applications open.

sarahboyce:
(need to set up the cron job and generally agree this is wanted)
image

### DIFF
--- a/home/management/commands/applications_open_email.py
+++ b/home/management/commands/applications_open_email.py
@@ -1,0 +1,71 @@
+from django.conf import settings
+from django.core.mail import send_mail
+from django.core.management.base import BaseCommand
+from django.template.loader import render_to_string
+from django.utils import timezone
+
+from accounts.models import CustomUser
+from home.models import Session
+
+
+class Command(BaseCommand):
+    help = """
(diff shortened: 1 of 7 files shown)


home/management/commands/applications_open_email.py
```diff
@@ -0,0 +1,71 @@
...
+        except Session.DoesNotExist:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    "There are no sessions with applications starting today"
+                )
+            )
+            return False
+        emails_sent = 0
+        for user in (
+            CustomUser.objects.select_related("profile")
```
tim-schilling, line 32: Do we need to select profile here? I don't see it being used in the formation of `email_data`.

tim-schilling (approved):
Looks good!

home/management/commands/applications_open_email.py
sarahboyce: `user.profile.create_unsubscribe_link()`

tim-schilling: D'oh. Thank you!

Issue #303, opened 2024-02-21
Djangonaut application process


sarahboyce:
We currently receive applications via google forms.
We like that we can quickly tweak wording and are flexible to make changes to the application.

We are discussing options to have this process integrated in the website, some options are:
- we build this ourself
- we use a package like https://pypi.org/project/django-form-surveys/
- we embed the google form in the site and extract the response after submission (we need to review them)

**Nice to have features for Djangonauts (submitters):**
- Djangonauts can view their past submissions (and so re-use some responses if they wish)
- Djangonauts can save a draft

**Nice to have features for Organizers (reviewers):**
- can see total number of submissions
- can add an internal "comment" and score on an answer (each question response can receive multiple comments and scores)
- can edit/delete these comments and scores

**Other notes**
We had a challenge for gaining availability and timezone information. 
- Timezone should be a dropdown of a standardised format
- Ideally we collect a time slots (like in calendly? we can embed a ready made solution) of when people are available (saved to the database in the same timezone) which we can use to help form groups with similar availabilities 

Some information that we collected in the application may exist or better belong in the user profile and we should make them fill this out before applying. E.g. name, preferred pronouns, GitHub username etc.

sarahboyce:
Update:
- We went for the option of building this ourselves
- We have done both of the 'other required features' (update profile and require email confirmation)
- We need to be able to see past submissions, this will be covered mostly by: https://github.com/djangonaut-space/wagtail-indymeet/issues/353
- We need the features for the organisers, this will be covered mostly by: https://github.com/djangonaut-space/wagtail-indymeet/issues/332

We have a rough test application on staging: https://staging-djangonaut-space.azurewebsites.net/survey_response/create/djangonaut-application-2024-session-2/
You need to be a user, who has not submitted, with a confirmed email to complete the survey. Having testers and feedback on the process would be great.

tim-schilling:
This is working reasonably well at this point. The next step is to identify what critical features are we missing.

- [ ] Create a survey that matches our Google form (example here) and note where there are deficiencies
- [ ] Check response length validations
- [ ] Check for user experience and email notification flows (I don't think we have email notifications)
- [ ] How are we handling timezones and availabilities?

PR #306, opened 2024-02-22
Add ability for user to answer dynamic surveys.

sarahboyce:
Start of https://github.com/djangonaut-space/wagtail-indymeet/issues/303

This is using https://github.com/irfanpule/django-form-surveys/blob/master/djf_surveys/models.py for inspiration

image

image


home/models/survey.py
```diff
@@ -0,0 +1,151 @@
...
+            update_fields = {"slug"}.union(update_fields)
+        super().save(
+            force_insert=force_insert,
+            force_update=force_update,
+            using=using,
+            update_fields=update_fields,
+        )
+
+
+class UserAnswer(BaseModel):
```
tim-schilling, line 112: If I'm being nitpicky, I'd ask for this to be called a `UserResponse` since it has nothing to do with `Answer`.

home/models/survey.py
```diff
@@ -0,0 +1,151 @@
...
+    user = models.ForeignKey("accounts.CustomUser", on_delete=models.CASCADE)
+
+    class Meta:
+        ordering = ["-updated_at"]
+
+    def __str__(self):
+        return str(self.id)
+
+
+class Answer(BaseModel):
```
tim-schilling, line 123: And then this could be called `UserAnswer` (I could see us eventually creating a generic `Answer` model).

home/models/survey.py
RachellCalhoun: Is this the SurveyResponse? Maybe similar naming would be helpful.

RachellCalhoun:
1. Do we imagine any conditional answers? 
2. Will these all be displayed on one page or in sequential pages? (in this case can they go back?)  
General comment about naming. 
Could we pull out the naming a bit and make it more standard, for example: `Question`, `QuestionResponse` would be possible response for that question. And then `UserResponse` would be the selected user response for that question. `SurveyResponse or `UserSurveyResponse` for the collection of answers for that survey instance for that user.

home/models/survey.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class TypeField(models.TextChoices):
+    TEXT = "TEXT", _("Text")
+    NUMBER = "NUMBER", _("Number")
+    DATE = "DATE", _("Date")
+    RADIO = "RADIO", _("Radio")
+    SELECT = "SELECT", _("Select")
+    MULTI_SELECT = "MULTI_SELECT", _("Multi Select")
+    TEXT_AREA = "TEXT_AREA", _("Text Aria")
```
RachellCalhoun, line 51: spelling area

Issue #301, opened 2024-02-19
Redirect to the referrer after login (rather than home)


sarahboyce:

raffaellasuardini:
Closed by: #309

PR #309, opened 2024-02-26
Redirect after login

raffaellasuardini:
## Description

After login there's a redirect to the previous page 

* redirect_field_name = 'next_page' because next is already used for other purpuse

## Issue

* https://github.com/djangonaut-space/wagtail-indymeet/issues/301


indymeet/settings/base.py
```diff
@@ -211,7 +211,7 @@
 
 AUTH_USER_MODEL = "accounts.CustomUser"
 
-LOGIN_REDIRECT_URL = "/"
+#LOGIN_REDIRECT_URL = "/"
```
tim-schilling, line 214: I believe this is an expected Django setting, so I'm not sure we want to remove it. Maybe it'd be best to go to the account/profile edit url as a fallback.

accounts/urls.py
```diff
@@ -7,8 +7,10 @@
 from .views import profile
 from .views import SignUpView
 from .views import unsubscribe
+from django.contrib.auth import views as auth_views
 
 urlpatterns = [
+    path('login/', auth_views.LoginView.as_view(redirect_field_name='next_page'), name='login'),
```
tim-schilling, line 13: Looks like you may not have `pre-commit` installed which is because it's not in the contributing docs :upside_down_face: [Install pre-commit on your machine](https://pre-commit.com/#install), then run `pre-commit install` in the project directory. It'll apply our style guidelines when you attempt to commit something.

For example, it would have automatically changed the `'` to `"`. That particular rule makes it easier to search for a string such as `"login"` because you don't have to search for both `"login"` and `'login'` in the codebase.

tim-schilling (changes requested):
Thank you for picking this one up @raffaellasuardini. It's a personal peeve of mine to be redirected to the marketing page on login rather than where ever I was or a user specific page.

accounts/urls.py
tim-schilling: I opened https://github.com/djangonaut-space/wagtail-indymeet/issues/310 to address the inadequate documentation

indymeet/settings/base.py
raffaellasuardini: I was not sure what to do in this scenario. In the documentation I read that the default value for `LOGIN_REDIRECT_URL` is of course `account/profile` but maybe it's better explicit the url? Do you think I just could delete the line and leave the [defaul value](https://docs.djangoproject.com/en/5.0/ref/settings/#login-redirect-url) instead?

accounts/urls.py
raffaellasuardini: I was a bit scared to use pre-commit because it's not mentiond in the docs. 
Is the use of `pre-commit run --all-files` not allowed, right?

indymeet/settings/base.py
tim-schilling: Sorry for the incredibly late reply. We do have a profile page, let's use the URL for that.

tim-schilling: It should be accounts/profile to point to this path https://github.com/djangonaut-space/wagtail-indymeet/blob/develop/accounts%2Furls.py#L14

accounts/urls.py
tim-schilling: That's my fault when I added it I didn't mention it. Don't worry about running it for all files. By default it's limited to the ones you're changing and that's good enough.

If we missed a number of files, that's something to be cleaned up in a separate PR

indymeet/settings/base.py
raffaellasuardini: To use the default value (the profile page) I will simply delete the variable. Do you agree?

tim-schilling: Yes, you're correct. I wasn't sure if the default value was `account/profile` or `accounts/profile`.

raffaellasuardini: Ok, I've done it. Could you check one last time @tim-schilling ?

tim-schilling (approved):
Looks great! Thank you

Issue #263, opened 2024-02-12
Style General Page wagtail type


tim-schilling:
The general page is currently unstyled with any of the tailwind styles. It can replicate the blog for now.

PR #314, opened 2024-03-13
Fix formatting issue on blog post

sabderemane:
Prior to this changes, the class prose was setting
a specific width and break the responsiveness with the blog container.

This change also add the class
`text-right` in order to align the text on the right side when the image is on the right class

Note: I didn't fully tested it, I haven't finished to configured again my setup


tim-schilling:
I'm fine with this being merged to test and iterate on immediately. Otherwise we should probably check how this changes the blog posts visually (and if it fixes #263)

PR #326, opened 2024-03-22
Update some blog styles and add docker setup

sarahboyce:


theme/static_src/src/djangonaut-space.css
```diff
@@ -45,8 +45,16 @@
...
+        padding: 20px 40px;
         margin-bottom: 20px;
+
+        p {
+            color: var(--dark);
+            margin: 0 0 1em 0;
+        }
+        a {
+            color: var(--ds-purple) !important;
+        }
```
sarahboyce, line 57: These changes didn't really work 😞 
@sabderemane I want to have the following changes in the blog:
- links be purple
- text be dark
- paragraphs have margin underneath

Do you know where I need to put this?

tim-schilling: Do we need to put this as a class on all the `a` elements?

tim-schilling (approved):
Looks generally good. I think your question for Sarah A isn't a blocker.

PR #334, opened 2024-03-26
Standardize the blocks we are using across blog and general pages

RachellCalhoun:
related to #263


tim-schilling (commented):
Love this. Thank you

tim-schilling:
Do we need to delete any existing content before deploying this?

RachellCalhoun:
@tim-schilling 
> Do we need to delete any existing content before deploying this?

None in prod, I copied the content then deleted two general pages in staging. So we should be good to go.

tim-schilling:
Cool, let's merge this then deploy to production with the other style change.

PR #335, opened 2024-03-26
Add profile/user update page.

sarahboyce:
image

Adds a way to update your profile.
Need to update the email confirmation to be more generic but basics are here.

### DIFF
--- a/accounts/views.py
+++ b/accounts/views.py
@@ -18,9 +18,9 @@ from django.utils.encoding import force_str
 from django.utils.http import urlsafe_base64_decode
 from django.utils.http import urlsafe_base64_encode
 from django.views import View
-from django.views.generic.edit import CreateView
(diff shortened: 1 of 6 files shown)


accounts/views.py
```diff
@@ -81,35 +105,69 @@ def form_valid(self, form):
...
+        Returns the initial data to use for forms on this view.
+        """
+        initial = super().get_initial()
+        initial["receive_program_updates"] = (
+            self.request.user.profile.receiving_program_updates
+        )
+        initial["receive_event_updates"] = (
+            self.request.user.profile.receiving_event_updates
+        )
+        initial["receiving_newsletter"] = self.request.user.profile.receiving_newsletter
```
RachellCalhoun, line 135: I think this needs to be `receive_newsletter`

RachellCalhoun (commented):
One inline comment but otherwise looks good.

Issue #337, opened 2024-03-27
Email confirmation flow - make visible, resend email option, permissions


sarahboyce:
We have a flow to confirm your email already. What we still need is:

- [x] a warning to a user that they have not yet confirmed their email
- [ ] a way to resend this email
- [x] limited permissions to the user when they haven't confirmed their email (e.g. they cannot see the 'Apply' buttons or access surveys)
- [ ] tests for all of this

sarahboyce:
We need an button to resend the confirmation email, probably on the profile page. Can also be on the warning. 
We need to update the email to make it more generic (as it is a thank you for signing up and maybe we make it more about email confirmation only).

PR #340, opened 2024-03-27
Add permission checking for session applications.

sarahboyce:
- When a user is logged in but hasn't confirmed their email address, they will now see a warning that they need to confirm this as they *may* not be able to apply to sessions without confirming their email (see screenshot)
- Sessions which do not have an application_url defined but instead have a application_survey will enforce that they cannot Apply (see the Apply button or access the survey) unless they are logged in and verified their email
- If they already applied for the session using the survey, the Apply button will no longer be visible

When we only use our internal survey's in our application, we can simplify the logic.

This addresses bullet point 1 and 3 in this issue: https://github.com/djangonaut-space/wagtail-indymeet/issues/337

image


tim-schilling:
@sarahboyce You'll need to pull the latest. I prepped it after we merged the large set of pre-commit changes

home/views.py
```diff
@@ -92,27 +91,35 @@ class SessionDetailView(DetailView):
...
+    def test_func(self):
         survey = self.get_object()
-        if UserSurveyResponse.objects.filter(survey=survey, user=request.user).exists():
-            messages.warning(request, gettext("You have already submitted."))
-            return redirect("session_list")
-        return super().dispatch(request, *args, **kwargs)
+        user = self.request.user
+        return (
+            user.profile.email_confirmed
+            and not UserSurveyResponse.objects.filter(survey=survey, user=user).exists()
```
tim-schilling, line 121: I understand why we need this, but I think it belongs in `CreateUserSurveyResposneForm.clean()` so it could be rendered in  [`non_field_errors`](https://docs.djangoproject.com/en/5.0/ref/forms/api/#django.forms.Form.non_field_errors). I think a 403  error (I'm assuing that's what UserPassesTestMixin returns) is less obvious to the user why their submission was rejected.

tim-schilling (approved):
This is good. I'm not a fan of the 403 error though, but this is a lot of steps forward and worth merging.

home/views.py
sarahboyce: Ah that would mean they fill out the form again and submit and learn they're not allowed?
I think eventually I'll have a way for them to see their submissions and it would be available to them

tim-schilling: Fair enough!

PR #343, opened 2024-03-28
Pre commit run all files

MHLut:
This is a separate fix for pre-commit issues, since the pre-commit CI will otherwise commit them to each pull request.

I could not fix every issues yet because some aren't as simple as they look, and also I haven't been able to run pytest locally to check if I hadn't broken anything.


MHLut:
@tim-schilling Heads-up

.git-blame-ignore-revs
```diff
@@ -1,3 +1,4 @@
 # .git-blame-ignore-revs
 # Apply pre-commit to the project
 b9f6c18a38a38728e27889762300d6a30821bfaa
+# ToDo: add Marijke Luttekes's pre-commit fix here once it's been merged to development
```
MHLut, line 4: This is for a separate PR after merging this one.

sarahboyce (approved):
💪

PR #349, opened 2024-03-29
Add basic load test utilities to the application.

tim-schilling:


sarahboyce (approved):
This is great! Ran it locally ⭐ thank you for setting it up

PR #350, opened 2024-03-29
Don't update wagtail reference index for survey responses

tim-schilling:
Follow-up to #349

### DIFF
--- a/home/models/survey.py
+++ b/home/models/survey.py
@@ -118,6 +118,11 @@ class Question(BaseModel):
 class UserSurveyResponse(BaseModel):
     survey = models.ForeignKey(Survey, on_delete=models.CASCADE)
     user = models.ForeignKey("accounts.CustomUser", on_delete=models.CASCADE)
+    # Wagtail has a post_save handler that runs anytime a model is saved
+    # that wagtail cares about to create a reference index. It doesn't matter
+    # in this context so let's ignore it.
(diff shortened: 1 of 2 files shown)


sarahboyce (approved):
whaaaaaaat 🤯 
This is great ⭐ 
Maybe we need this for some of the account models also? (maybe profile)

tim-schilling:
We may need to find a better way to do this in our app. I'm also not sure how it all works, but this is one option available to us.

sarahboyce:
I think this problem goes away when we upgrade to wagtail 5+
https://docs.wagtail.org/en/v5.1.3/advanced_topics/reference_index.html

tim-schilling:
Good find.

tim-schilling:
Solved in #360

tim-schilling:
@sarahboyce just read the docs link. I'm not sure I agree. Can you highlight the part that says that this is solved?

sarahboyce:
"When introduced in Wagtail 4.1, the Reference Index recorded references in all application models by default. Wagtail 5.0 reduced the scope of the default models to those specifically linked with Wagtail."
https://docs.wagtail.org/en/v5.1.3/advanced_topics/reference_index.html#configuration

tim-schilling:
Thank you. I'll need to rerun this test without the ignore prop to confirm then.

PR #352, opened 2024-03-29
Register UserQuestionResponse and UserSurveyResponse in the admin.

### LINKED ISSUE #332: Need survey responses in the admin
We need to expose `UserQuestionResponse` and `UserSurveyResponse` in the [django admin](https://github.com/djangonaut-space/wagtail-indymeet/blob/develop/home/admin.py). 

related to: https://github.com/djangonaut-space/wagtail-indymeet/issues/303

Pradhvan:
Add `UserQuestionResponse` and `UserSurveyResponse` in `home/admin.py`

Closes #332

### DIFF
--- a/home/admin.py
+++ b/home/admin.py
@@ -1,13 +1,15 @@
 from django.contrib import admin
 from django.urls import reverse
-from django.utils.html import escape
 from django.utils.safestring import mark_safe
 
-from .models import Event, ResourceLink
+from .models import Event
+from .models import ResourceLink
 from .models import Question
 from .models import Session
 from .models import SessionMembership
 from .models import Survey
+from .models import UserQuestionResponse
+from .models import UserSurveyResponse
 
 
 @admin.register(Event)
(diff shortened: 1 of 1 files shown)


Pradhvan:
I have just added models with all the fields and no filters. Please let me know if those are needed here. Will modify the PR. :)

home/admin.py
```diff
@@ -56,3 +58,13 @@ class SurveyAdmin(admin.ModelAdmin):
...
+
+
+@admin.register(UserQuestionResponse)
+class UserQuestionResponseAdmin(admin.ModelAdmin):
+    model = UserQuestionResponse
+
+
+@admin.register(UserSurveyResponse)
+class UserSurveyResponse(admin.ModelAdmin):
+    model = UserSurveyResponse
```
sarahboyce, line 70: Add a filter by survey 👍

sarahboyce:
Thank you for picking this up!

Would love some screenshots ❤️ 
I think both of these shouldn't be edittable and we _maybe_ shouldn't be able to add entries

tim-schilling:
Let's go ahead and use raw id fields or autocomplete fields for the user relationships ( if they exist). Dropdowns won't work with more than 500ish results.

PR #354, opened 2024-03-29
Lazy load wagtail images.

sarahboyce:
https://docs.wagtail.org/en/latest/advanced_topics/performance.html#image-attributes


tim-schilling (approved):
This will help on the browser side, but I don't think it'll improve the server's render time which is where the major bottleneck currently is.

PR #367, opened 2024-03-31
Prod deploy - updated auth views and confirm before unsubscribe

sarahboyce:
Tested on staging 👍


tim-schilling (approved):
Looks good. I didn't test anything, just looked over the code. Seems reasonable

Issue #353, opened 2024-03-29
Add a UserSurveyResponseDetailView so users can see their survey submission


sarahboyce:
This will have the survey details and questions and the answers from the user.
This should only be viewable by the user themselves for now.
Have this linked to the session list view and detail view cards when it has a application_survey defined

PR #370, opened 2024-03-31
Add a user survey response view.

sarahboyce:
Fixes https://github.com/djangonaut-space/wagtail-indymeet/issues/353

image

### DIFF
--- a/home/managers.py
+++ b/home/managers.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-from django.db.models import Exists
 from django.db.models import OuterRef
+from django.db.models import Subquery
 from django.db.models import Value
 from django.db.models.query import QuerySet
 from django.utils import timezone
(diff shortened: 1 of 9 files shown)


home/views.py
```diff
@@ -143,3 +144,30 @@ def post(self, request, *args, **kwargs):
...
+        user = self.request.user
+        return user == survey_response.user
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs["user_survey_response"] = self.get_object()
+        return kwargs
+
+    def get_context_data(self, **kwargs):
+        survey = self.get_object().survey
```
tim-schilling, line 168: I don't think it's a terrible idea to select related and prefetch related all the relevant data that the form is going to use.

home/managers.py
```diff
@@ -37,10 +37,10 @@ def with_applications(self, user):
         if user.is_anonymous:
             return self.annotate(completed_application=Value(False))
         return self.annotate(
-            completed_application=Exists(
+            completed_application=Subquery(
```
tim-schilling, line 40: Where is this change being used?

sarahboyce: It's used in `home/templates/home/includes/session_apply_btn.html` to have the id in order to create the link to the survey submission

tim-schilling (approved):

PR #379, opened 2024-04-02
375 Center Event and Session Detail View

### LINKED ISSUE #375: Update style/formatting on events and session pages
These are full width size for these pages, we should update them so they are more similar to our general pages and blog pages. The template needs to be updated. 

I think we could probably use a base template to make it easier to standardize General and Session / Events pages. 

For reference this is what was updated on the general page: https://github.com/djangonaut-space/wagtail-indymeet/pull/264/files

RachellCalhoun:
closes #375

### DIFF
--- a/indymeet/templates/puput/entry_page.html
+++ b/indymeet/templates/puput/entry_page.html
@@ -43,14 +43,14 @@
 
 {% block content %}
         <div class="container mx-auto flex justify-center">
-            <article class="box max-w-4xl blog-container"
+            <article class="box max-w-5xl blog-container"
                     {% if self.id %}data-entry-page-update-comments-url="{% url 'entry_page_update_comments' self.id %}{% endif %}">
                 {% include 'puput/entry_page_header.html' with entry=self %}
                 <section>
                     {% if self.header_image %}
                         <div class="image featured">
                             {% image self.header_image fill-800x450 as header_image %}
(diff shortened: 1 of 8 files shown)


RachellCalhoun:
image
image

sarahboyce (approved):
We should probably do this for some of the templates in `indymeet/templates/registration` too 👍

RachellCalhoun:
> We should probably do this for some of the templates in `indymeet/templates/registration` too 👍

I agree, let me spend some more time on this branch with the sessions/events and try to clean up /standardize some of these templates a bit more.

indymeet/templates/puput/entry_page.html
```diff
@@ -43,14 +43,14 @@
 
 {% block content %}
         <div class="container mx-auto flex justify-center">
-            <article class="box max-w-4xl blog-container"
+            <article class="box max-w-5xl blog-container"
```
RachellCalhoun, line 46: if we wanted it the same as the rest of the pages, we'd have w-3/4 instead of max-w-5xl, but it feels too big here especially with our smaller font etc.

RachellCalhoun:
All pages should adhere to the new styling: 
image

sarahboyce (approved):
Thank you for doing this 🌟

Issue #372, opened 2024-04-01
Update image blocks to allow links


RachellCalhoun:
We want to 

- [ ] add optional link to any block with an image. We can probably use URLBlock (https://docs.wagtail.org/en/latest/reference/streamfield/blocks.html#wagtail.blocks.URLBlock)
- [ ] update image template to include link if one exists 
- [ ] make sure formatting will work with various images (rect/square)  

One example: 
`RightImageLeftTextBlock` https://github.com/djangonaut-space/wagtail-indymeet/blob/develop/home/blocks.py#L109-L118
template: https://github.com/djangonaut-space/wagtail-indymeet/blob/develop/indymeet/templates/blocks/right-image-left-text.html

To see these changes reflected you will need to make migrations, migrate, then you can create a new page (GeneralPage) and select the updated block and you should get a link option.

RachellCalhoun:
The main driver for this was the sponsors -- but maybe we use the cards above, or create a logo block. It depends on how we want to display it. I think having a logo block would be helpful. https://tailwindcomponents.com/component/tailwind-css-logo-area-with-title

PR #381, opened 2024-04-02
372 Add Image Card Blocks

RachellCalhoun:
addresses #372 

I went with creating a new feature instead of updating all existing. 
I also increased max-width on our container by one size.


RachellCalhoun:
image

sarahboyce (approved):
Looks really nice 😍 
Migrations might need to be regenerated depending the merge order

Issue #245, opened 2024-01-23
Placeholder session


dawnwages:
if none open, then show placeholder

dawnwages:
https://djangonaut.space/sessions/

tim-schilling:
Completed in #392

PR #392, opened 2024-04-16
Feature: placeholder for sessions

raffaellasuardini:
## Description
I used the method of Session model `is_accepting_applications` to check if a session is open for the application.  Is it better to check the end of the `end_date` which is the end of the 8-week program? 

## Issue:

* #245

### DIFF
--- a/home/templates/home/prerelease/session_list.html
+++ b/home/templates/home/prerelease/session_list.html
@@ -25,7 +25,17 @@
         </div>
         <div class="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 pt-4">
             {% for session in sessions %}
+                {% if session.is_accepting_applications %}
                 {% include 'home/includes/session_card.html' %}
+                {% endif %}
+                {% if forloop.last and not session.is_accepting_applications %}
+                     <div class="col col-lg-3 col-md-4 col-sm-6 col-12">
+                        <div class="max-w-sm rounded overflow-hidden shadow-lg card">
+                          <div class="px-6 py-4 card-body">
+                              <h5 class="font-bold text-xl mb-1 card-title">No active session</h5>
+                                <p class="text-gray-700 text-base card-text">While we're busy checking our on-board instrumental code for the next cosmic journey, you can prepare by exploring our Program Documentation <a href="https://github.com/djangonaut-space/program">(https://github.com/djang…)</a>.</p>
+
+                {% endif%}
             {% endfor %}
         </div>
     </div>
--- a/home/templates/home/prerelease/session_list.html
+++ b/home/templates/home/prerelease/session_list.html
@@ -33,7 +33,7 @@
                         <div class="max-w-sm rounded overflow-hidden shadow-lg card">
                           <div class="px-6 py-4 card-body">
                               <h5 class="font-bold text-xl mb-1 card-title">No active session</h5>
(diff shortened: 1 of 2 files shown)


home/templates/home/prerelease/session_list.html
```diff
@@ -25,7 +25,17 @@ <h1 class="text-5xl">Sessions</h1>
         </div>
         <div class="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 pt-4">
             {% for session in sessions %}
+                {% if session.is_accepting_applications %}
```
tim-schilling, line 28: I'm not sure we want to hide past sessions. Though I do think we want these in the opposite order so that the first elements are the most recent.

home/templates/home/prerelease/session_list.html
```diff
@@ -25,7 +25,17 @@ <h1 class="text-5xl">Sessions</h1>
...
                 {% include 'home/includes/session_card.html' %}
+                {% endif %}
+                {% if forloop.last and not session.is_accepting_applications %}
```
tim-schilling, line 31: I agree with your idea of checking to see if the end_date is in the future. We only want to show this message if we have no session in the future.

home/templates/home/prerelease/session_list.html
```diff
@@ -25,7 +25,17 @@ <h1 class="text-5xl">Sessions</h1>
...
+                     <div class="col col-lg-3 col-md-4 col-sm-6 col-12">
+                        <div class="max-w-sm rounded overflow-hidden shadow-lg card">
+                          <div class="px-6 py-4 card-body">
+                              <h5 class="font-bold text-xl mb-1 card-title">No active session</h5>
+                                <p class="text-gray-700 text-base card-text">While we're busy checking our on-board instrumental code for the next cosmic journey, you can prepare by exploring our Program Documentation <a href="https://github.com/djangonaut-space/program">(https://github.com/djang…)</a>.</p>
```
tim-schilling, line 36:
```suggestion
                                <p class="text-gray-700 text-base card-text">While we're busy checking our on-board instrumental code for the next cosmic journey, you can prepare by exploring our <a href="https://github.com/djangonaut-space/program">Program Documentation</a>.</p>
```

home/templates/home/prerelease/session_list.html
raffaellasuardini: I'm not sure. How can I check if there's no session in the future? Is it ok to check the end of the program? For example for session2 the end is 12th August, if there are no other sessions available on 13th August then I display the placeholder, is it ok?

home/templates/home/prerelease/session_list.html
raffaellasuardini: Is the placeholder will be displayed at the beginning of the other sessions?

tim-schilling: Yes I think that makes the most sense.

home/templates/home/prerelease/session_list.html
tim-schilling: Yes, you're correct. We would make the assumption that the first session will be the newest (the view needs to be updated to do this). It could potentially even be in the future. Or it could be in the past. When it's in the past, that's when we want to show the placeholder.

tim-schilling:
@raffaellasuardini is this something you're still working on?

raffaellasuardini:
@tim-schilling Yes! sorry for the delay. I'm going to push my updates.

raffaellasuardini:
@tim-schilling what do you think?

tim-schilling (approved):
Looks good! Thank you

PR #436, opened 2024-12-11
Fixed the responsiveness of the landing page

raffaellasuardini:
I adjusted the custom CSS classes to get a **sticky footer**  and to use breakpoints in the body. 
I will attach some screenshots for reference

Screenshot 2024-12-11 alle 19 45 00

Screenshot 2024-12-11 alle 19 46 19

### DIFF
--- a/theme/static_src/src/djangonaut-space.css
+++ b/theme/static_src/src/djangonaut-space.css
@@ -79,14 +79,14 @@
 ----------------------------------------------------------------------*/
 
 .landing {
-    max-width: 100vw;
+    /*max-width: 100vw;*/
     min-height: 100vh;
     min-height: 100vh;
     margin: 0 auto;
     padding: 0 15px;
     display: flex;
     flex-direction: column;
-    justify-content: center;
+    /*justify-content: center;*/
     align-items: center;
     background: black;
     color: white;
@@ -118,9 +118,9 @@
 }
 
 .landing-page-footer {
-    position: fixed;
+   /* position: fixed;
     bottom: 0;
-    left: 0;
+    left: 0;*/
     display: flex;
     align-items: center;
     flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


theme/static_src/src/djangonaut-space.css
```diff
@@ -79,14 +79,14 @@
...
-    max-width: 100vw;
+    /*max-width: 100vw;*/
     min-height: 100vh;
     min-height: 100vh;
     margin: 0 auto;
     padding: 0 15px;
     display: flex;
     flex-direction: column;
-    justify-content: center;
+    /*justify-content: center;*/
```
tim-schilling, line 89: Is there a reason why these are commented out rather than deleted?

raffaellasuardini: I wasn't sure if it was the right file to update.

tim-schilling: Gotcha. I think we need to decide if it is and then either remove or add the code. Commented out code can be confusing to people, especially when debugging a problem. If I do need to commit commented or code, I like to leave an explicit reason why it's commented out and when it should be uncommented or removed.

raffaellasuardini: Ok, in this case, I will remove `justify-content: center` because I use the breakpoints in the body class. I don't find any other reference for this class.

theme/static_src/src/djangonaut-space.css
```diff
@@ -118,9 +118,9 @@
 }
 
 .landing-page-footer {
-    position: fixed;
+   /* position: fixed;
```
raffaellasuardini, line 121: in this case, I will remove the position fixed to use `grow` on `main` to achieve the sticky footer

tim-schilling (approved):
Thank you @raffaellasuardini!

Issue #433, opened 2024-12-05
Add BlueSky account to socials on site 


RachellCalhoun:
I know these appear in the footer, but possibly elsewhere? Anyway, we should add it in. https://bsky.app/profile/djangonaut.space

tim-schilling:
The landing page

raffaellasuardini:
This is closed by #437

PR #437, opened 2024-12-12
Added Bluesky iconto to the footer of landing page. 

raffaellasuardini:
I added Bluesky to the landing page footer. 
Since the Bluesky brand icon is only [available from `6.5.2 `or `6.7.1`](https://fontawesome.com/v6/icons/bluesky?f=brands&amp;s=solid), I also updated the local Fontawesome and the CDN.

I found other references to some icons, but in the blog session, that is linked to Puput and Fontawesome 4, so I didn't touch them.

This solves #433 

Happy to add the Bluesky reference in the non-landing page footer if needed


tim-schilling:
> Happy to add the Bluesky reference in the non-landing page footer if needed

Do you think we need it there? I'm not 100% sure.

raffaellasuardini:
> > Happy to add the Bluesky reference in the non-landing page footer if needed
> 
> Do you think we need it there? I'm not 100% sure.

Neither do I, and there's no LinkedIn or GitHub. Mastodon, YouTube, and X may be more relevant and important to the home page because there is a lot of content there.

tim-schilling:
Hmm. I wonder if we should switch to the icons rather than full links? That way it's easier to add as many as we want.

raffaellasuardini:
Is there a specific reason why there are long links? Otherwise, this could be an option

tim-schilling:
Not that I'm aware of.

raffaellasuardini:
if that works for you I can push it into this PR

raffaellasuardini:
Just an idea, but I was thinking of making the icons the same color as ds-purple.
Screenshot 2024-12-13 alle 15 55 54

tim-schilling:
Love that!

PR #442, opened 2025-01-07
Added support for ResourceLink model and view.

### LINKED ISSUE #441: Support resource links
We utilize Google docs pretty heavily and need to copy resources from session to session. One problem is that sometimes we need to clear the history of a document which can only be done by copying the Google doc. This creates a new shareable link for it.

We need a model that takes a path and redirects to a fully qualified URL.

The goal is to have urls for the public workbook templates

tim-schilling:
Closes #441


home/models/resource.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ResourceLink(models.Model):
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
+    permanent = models.BooleanField(default=False)
+    path = models.TextField(
+        help_text="The relative path for requests to the djangonaut.space web app.",
+        unique=True,
+    )
```
raffaellasuardini, line 11: Have you already discarded the possibility of `SlugField` or could it be used?

tim-schilling: A `SlugField` doesn't allow for the `/` character to support paths so we can't use that directly. We could use a similar validator, but I'm not sure it buys us a ton. Our use-case is pretty minimal.

RachellCalhoun (approved):
lgtm! :rocket: Thanks!

PR #450, opened 2025-01-23
Fixed social icons inside blog

raffaellasuardini:
I removed fontawesome old reference from Puput.
Updated icons inside the blog

I also found 'user' and 'calendar' icons, but I wasn't able to update them because they're probably used inside the Wagtail.

This should fix #445

PR #453, opened 2025-01-28
Production Release - 2025-01-28

tim-schilling:
PRs:
- #450


tim-schilling:
I suppose I should wait until after applications close. It seems irresponsible to do this when people may actually need the site.

tim-schilling:
This release never occurred.

PR #454, opened 2025-01-28
Changed default puput blog main color to purple.

tim-schilling:


raffaellasuardini:
Sorry @tim-schilling for the long long waiting, and thank you for helping me fixing my previous bug.
I'm not quite familiar with wagtail, so I have a question.
You created an `EntryAbstract`, but I saw that's already present `BlogAbstract` in the `blog.py` model here: 
[https://github.com/djangonaut-space/wagtail-indymeet/blob/ee8fb2104e871a10d745799cdb43486ff5c9c464/home/models/blog.py#L42](URL)
Are there any reasons to separate `blog.py` and `puput_abstracts.py`?

tim-schilling:
> Are there any reasons to separate blog.py and puput_abstracts.py?

That's a good question. I'm not 100% sure on what's the best practice, but the justification in mind is that those puput abstract models are not actual models. Because of that, we should put them in a separate file. That helps communicate these exist for a particular reason that's related to puput.

> You created an EntryAbstract, but I saw that's already present BlogAbstract in the blog.py model here:

Not sure if you're asking why I'm renaming an abstract when one already exists. The reason is that the previous name was actually associated with the wrong base abstract model. I've now created both with the proper names. I hope.

raffaellasuardini:
> That's a good question. I'm not 100% sure on what's the best practice, but the justification in mind is that those puput abstract models are not actual models. Because of that, we should put them in a separate file. That helps communicate these exist for a particular reason that's related to puput.

I agree to separate puput from the rest. Nice and neat.  

Can we delete the previous abstract `BlogAbstract` in `blog.py` now that we have new abstracts?

tim-schilling:
D'oh. I forgot that bit. Yes we should.

tim-schilling:
Thank you!

I think I'd like to hold off on merging changes until we get the site migrated to its new home.

raffaellasuardini:
Oh sure!
Let's save that for later. I'm happy this PR is completed.

PR #464, opened 2025-02-21
Switch to Dokku for deployments on a VPS

### LINKED ISSUE #463: Change hosting to alternative to Azure
We are looking to move our hosting off of Azure to another platform.

One option considered is Hetzner with Dokku. This would provide very low cost hosting and a reasonable deployment interface. The downside is that we'd have to manage the server entirely, including the database. Given the low pressure to maintain uptime, this seems like a good trade-off.

tim-schilling:
Closes #463

Remaining TODOs:
- [x] Set up CI to replace Azure CD
- [ ] Transfer the Hetzner account to contact@
- [x] Configure backups on Hetzner

Long term, we could set up a process to export database or instance backups to a S3 bucket or something similar.


tim-schilling:
I added contact@ as an admin on the hetzner project. To transfer things, I'll need to create a new project with the contact@ account, then transfer the server to it. I'm going to finish getting this setup and put that off for the future.

PR #467, opened 2025-04-01
Add the site to the django webring.

tim-schilling:
See: https://djangowebring.com

#### Screenshots to view the layout

Screenshot from 2025-04-01 16-59-48
Screenshot from 2025-04-01 17-00-08
Screenshot from 2025-04-01 17-00-28

### DIFF
--- a/home/templates/home/home_page.html
+++ b/home/templates/home/home_page.html
@@ -48,5 +48,9 @@
         </a>
     </div>
 </footer>
+<div class="z-10">
+    <webring-css site="https://djangonaut.space"></webring-css>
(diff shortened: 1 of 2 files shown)


tim-schilling:
I don't think Jeff intends to keep this up :sweat_smile:

jefftriplett:
@tim-schilling I wasn't planning on taking it down. I was surprised someone called my bluff and asked to be added to it. 

I will probably take Django Packages out of it, and maybe the newsletter because we can't add html to the footer.

tim-schilling:
In that case!

Issue #70, opened 2023-01-20
Add video link field for Event 


RachellCalhoun:
We will want to record and link the recording back to the Event. It should also be displayed on the event detail page localhost/events/2

PR #474, opened 2025-05-01
added video link

FarhanAliRaza:
fixes https://github.com/djangonaut-space/wagtail-indymeet/issues/70


home/models/event.py
```diff
@@ -61,7 +61,7 @@ class Event(ClusterableModel):
         related_name="events",
         on_delete=models.SET_NULL,
     )
-
+    video_link = models.URLField(blank=True, null=True)
```
tim-schilling, line 64: I think it's best if we don't use `null=True` and let the empty string count as the false-y value. We don't have many events, so locking the table to set existing rows to `""` won't be painful.

tim-schilling (approved):
Thanks @FarhanAliRaza!

PR #481, opened 2025-08-19
Remove CCBV from contribution opportunities

LilyAcorn:


tim-schilling:
Removed because the project is not accepting contributions at the moment.

PR #499, opened 2025-09-30
Remove unused templates and move prerelease template up

tim-schilling:
This removes unused templates and moves the "prerelease" templates up into the home templates directory.


FarhanAliRaza:
I am getting the error at the home URL 
```
TemplateDoesNotExist at /
home/home_page.html
```
home_page.html was deleted in this commit.

Issue #617, opened 2025-12-09
Application review action list


tim-schilling:
Rachell and I worked through the new application flow and identified the following items:

- [x] Session Membership should have a search for users
- [x] Switch SessionAdmin available projects to filter_horizontal
- [x] Update navigator and captain workbooks to mention setting availability
- [x] Check on email about creating a survey response
- [x] "Response sent!" should tell the user to set their availability
- [x] Single scorer is only showing 1 question. Needs to be per question.
- [x] Update the application form to include avaiability mention
- [x] Move the dropdown links to columns on the table.
- [x] Remove availability questions from captains and navigators forms
- [x] move survey question order field to right after the question label on survey detail edit page
- [x] add project link to the application's project list 

Remaining TODOs:
- [x] Write buttondown instructions for week -5
- [x] Backfill the waitlist on production
- [x] Run load test
- [x] Document waitlist flow
- [x] Go through the application flow with the website team as testing.
- [x] Discuss about permissions in the admin.
- [x] Update the email confirmation to include links to set availability and any sessions accepting applications

tim-schilling:
This has been completed.

PR #620, opened 2025-12-11
Limit session organizer permissions and data access in the admin.

tim-schilling:
References #617

@RachellCalhoun this introduces some limits on what data a non-superuser in the admin can access. Any user who is an Admin in the Djangonaut Space community should be a session organizer. Nobody else should be a superuser in the org because it will grant access to all session application data.

Session organizers will have read access to all session data for the sessions they are associated with (they have a SessionMembership instance with the role of organizer). These users should also have:

- `CustomUser.is_staff = True`
- Be a member of the `auth.Group` Session Organizers

These two controls will give the user the ability to access the admin and the relevant pages within it respectively. The data that is shown will depend on the SessionMembership model.

## On Granting access

It's been built so that when a user is added as a organizer on the Session edit view, it will automatically set `is_staff` and associate the user with the Session Organizers group. For now, any session organizer can make another person a session organizer. I looked into limiting that, but it wasn't easy so I punted. This bit of automation should make the process seamless.

After the session concludes, an Djangonaut Space Admin (someone with superuser access), will need to review the group's membership and revoke people's access.

---

If you're in agreement with the above, I can merge the code. If you spot flaws in the process, logic, or want to discuss something, please let me know.

### DIFF
--- a/accounts/management/commands/setup_session_organizers_group.py
+++ b/accounts/management/commands/setup_session_organizers_group.py
@@ -5,6 +5,8 @@ Management command to create or update the Session Organizers group with require
 from django.contrib.auth.models import Group, Permission
 from django.contrib.contenttypes.models import ContentType
 from django.core.management import BaseCommand
+from django.db import transaction
+from django.db.models import Q
 
 from accounts.models import CustomUser, UserAvailability, UserProfile
 from home.models import (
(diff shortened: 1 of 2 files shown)


tim-schilling:
@RachellCalhoun it wouldn't be right now. My friend Claude tried building that out from the start and I removed it :sweat_smile: Do you think we should re-add it? I suppose it does eliminate another thing we have to think about.

Issue #644, opened 2026-01-16
Limit survey responses based on session dates


tim-schilling:
When a Survey is set as a `Session.application_survey`, it should only accept responses (and edited responses) when the current date is between the application dates. I've been able to submit application responses before the session starts. I didn't check to see if it works afterwards, but I suspect it's broken as well.

To reproduce:
1. Create a session at http://127.0.0.1:8000/django-admin/home/session/
2. Set the application date to be in the future
3. Create a survey and associate it with that session http://127.0.0.1:8000/django-admin/home/survey/
4. Edit the session and set the application_survey to the newly created survey
5. Use the link column from https://djangonaut.space/django-admin/home/survey/ to try to submit a response
6. This should show a message that this application will be open on X date
7. Set the application dates to be in the past
8. Try to browse to the survey
9. This should show a message that this application is closed

tim-schilling:
Closed in #645

PR #645, opened 2026-01-17
Enhance survey form handling by adding application status checks. 

FarhanAliRaza:
Users are now informed if the survey is not yet open or closed, preventing submission during these periods. Update tests to verify access restrictions based on application dates.

Related to https://github.com/djangonaut-space/wagtail-indymeet/issues/644

image

### DIFF
--- a/home/tests/test_user_survey_response_form_views.py
+++ b/home/tests/test_user_survey_response_form_views.py
@@ -275,7 +275,7 @@ class EditUserSurveyResponseViewTests(TestCase):
             data={f"field_survey_{self.question.id}": "Excellent"},
             follow=True,
         )
-        self.assertContains(response, "You are no longer able to edit this.")
+        self.assertContains(response, "This application is closed.")
 
     def test_can_edit_session_application_within_deadline(self):
         # Create a session with active application period
--- a/home/views/surveys.py
+++ b/home/views/surveys.py
@@ -16,8 +16,30 @@ from home.forms import CreateUserSurveyResponseForm, EditUserSurveyResponseForm
 from home.models import Survey, UserQuestionResponse, UserSurveyResponse
 
 
+class ApplicationStatusMixin:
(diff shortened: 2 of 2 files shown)


home/views/surveys.py
```diff
@@ -69,10 +69,28 @@ def get_form_kwargs(self) -> dict:
         kwargs["user"] = self.request.user
         return kwargs
 
+    def get_application_status(self, survey):
+        """Check the survey's application status.
+
+        Returns (status, start_date) where status is 'open', 'not_yet_open', or 'closed'.
+        """
+        if (session := getattr(survey, "application_session", None)) is None:
+            return "open", None
```
tim-schilling: I think removing the is none check and indenting the remainder would remove one of the return statements making this easier to test and debug.

home/tests/test_user_survey_response_form_views.py
```diff
@@ -26,6 +28,34 @@ def setUpTestData(cls):
             label="How are you?",
         )
 
+    def test_cannot_access_survey_before_application_opens(self):
```
tim-schilling: Is there an existing test for when a session is configured and the date I'd valid? If not, it may be helpful to add a test for that scenario.

tim-schilling (changes requested):
Looks pretty good! Marking as requesting changes specifically because of the edit flow. The other two comments are judgement calls.

I'll want to look at the visuals before merging unless you want to put screenshots up. Not a problem either way.

home/views/surveys.py
FarhanAliRaza: done

home/tests/test_user_survey_response_form_views.py
tim-schilling: Ah, I didn't see `test_can_edit_session_application_within_deadline`. That covers what I was expecting. Marking as resolved.

tim-schilling (approved):

PR #690, opened 2026-02-15
feat: team formation view improvements

### LINKED ISSUE #679: Update the form teams view functionality
- The Check Overlap feature for navigators should check the availability with the existing Djangonauts, not just the captain.
- The team view should have a "Compare team's availability" or something similar that uses the compare availability page
- The results of the check navigator overlap should have a link to the compare availability page
- Support filtering on team preferences
- Assigning a person to a team that isn't their preference correctly fails, but doesn't present a message to the user

JohananOppongAmoateng:
Updates the Team Formation view to improve the accuracy of availability overlap checks, adds direct links to compare availability for teams and applicants, and introduces filtering by project preference. Also improves error feedback when assigning applicants to teams with conflicting preferences. Closes #679


home/forms.py
```diff
@@ -910,21 +911,41 @@ def calculate_navigator_overlap_context(self) -> dict:
         """
         team = self.cleaned_data["team"]
         navigators = self.get_team_navigators()
+
+        # Also get existing djangonauts on the team
+        djangonaut_memberships = (
+            SessionMembership.objects.for_team(team)
+            .filter(role=SessionMembership.DJANGONAUT)
```
tim-schilling: There's an queryset method that can be used here instead. `.djangonauts()`

home/templates/admin/partials/overlap_result_item.html
```diff
@@ -16,12 +16,18 @@
...
+        {% endif %}
     {% else %}
         {# Full format for group results #}
         <strong>{{ title }}</strong> (Team: {{ team.name }})<br>
         {% if navigators %}
             <em>{% trans "Navigators:" %} {% for nav in navigators %}{{ nav.get_full_name }}{% if not forloop.last %}, {% endif %}{% endfor %}</em><br>
         {% endif %}
+        {% if existing_djangonauts %}
+            <em>{% trans "Existing Djangonauts:" %} {% for d in existing_djangonauts %}{{ d.get_full_name }}{% if not forloop.last %}, {% endif %}{% endfor %}</em><br>
+        {% endif %}
```
tim-schilling: I don't think this is necessary. The side view on the page shows the existing djangonauts already.

home/tests/test_team_features_update.py
tim-schilling: I think these tests should go into existing test files or be named for the components they are testing. This filename is relevant only for this PR, but will be confusing outside of that context. Does that make sense?

home/views/team_formation.py
```diff
@@ -377,6 +385,7 @@ def get_teams_with_statistics(session: Session) -> list[TeamStatistics]:
                 navigator_meeting_hours=overlap_stats["navigator_meeting_hours"],
                 captain_meetings=overlap_stats.get("captain_meetings", []),
                 is_valid=overlap_stats["is_valid"],
+                compare_availability_url=compare_url,
```
tim-schilling: Since the `TeamStatistics` is being passed all this information in, it seems reasonable to have it do the URL generation rather than generating the URL and passing it in.

tim-schilling (changes requested):
Thank you for working on this @JohananOppongAmoateng! I think we can move some logic around to make things a bit cleaner and easier to maintain for the future. Let me know if you would like any clarifications.

tim-schilling (approved):
Thank you @JohananOppongAmoateng!

PR #699, opened 2026-02-26
feat: automate creating community events

### LINKED ISSUE #698: Automate creating community wide events
We have several events that we create which take up a fair amount of time because we need to create them on Zoom, then on Discord, then email them out to people. We also want to avoid allowing community members to view each other's email addresses when possible.

I think we can rework the Event logic in the application with the following features:
- Allow public or private event (public is posted on the website, private is just for Djangonaut Space users who have a Session Membership record)
- Allow linking to a session
- Allow copying a past event
- Should create a Zoom event using the meeting template or one that mutes people on entry and enables a waiting room
- Should create a Discord event
- Should send an email with a calendar event attachment to all members of the session or community (depending on private/public and session selection)
- Should be able to add extra emails to be included on the calendar invite (can be used for guest speakers), default to sessions@djangonaut.space

The discord integration may challenging. If it is, we can put that off and have the event created view, provide guidance to walk the user through creating the discord event step by step.

JohananOppongAmoateng:
Closes #698


JohananOppongAmoateng:
@tim-schilling  could you glance through and see if we on the right track.

home/email.py
```diff
@@ -35,10 +53,23 @@ def send(email_template, recipient_list, context=None, from_email=None):
...
+        msg = EmailMultiAlternatives(
+            subject=subject,
+            body=text,
+            from_email=from_email or settings.DEFAULT_FROM_EMAIL,
+            to=recipient_list,
+        )
+        msg.attach_alternative(html, "text/html")
+        for filename, content, mimetype in attachments:
+            msg.attach(filename, content, mimetype)
+        msg.send()
```
tim-schilling: It's probably fine to just use this logic rather than `send_mail()`.

home/models/event.py
```diff
@@ -62,6 +79,16 @@ class Event(ClusterableModel):
         on_delete=models.SET_NULL,
     )
     video_link = models.URLField(blank=True, default="")
+    is_public = models.BooleanField(default=True)
+    extra_emails = models.JSONField(
```
tim-schilling: Is there a reason you chose not to use an [`ArrayField`](https://docs.djangoproject.com/en/6.0/ref/contrib/postgres/fields/#arrayfield)?

home/templates/email/event_calendar_invite/body.html
```diff
@@ -0,0 +1,38 @@
...
+  {% if event.description %}
+  <p style="font-family: Helvetica, sans-serif; font-size: 16px; font-weight: normal; margin: 0; margin-bottom: 16px;">
+    {{ event.description|linebreaksbr }}
+  </p>
+  {% endif %}
+
+  <p style="font-family: Helvetica, sans-serif; font-size: 16px; font-weight: normal; margin: 0; margin-bottom: 16px;">
+    <strong>Location:</strong> {{ event.location }}<br>
+    <strong>Start:</strong> {{ event.start_time|date:"l, F j, Y g:i A e" }}<br>
+    <strong>End:</strong> {{ event.end_time|date:"l, F j, Y g:i A e" }}
```
tim-schilling: These should use the timeis custom template filter to to include links to time.is for easier timezone conversion.

home/views/events.py
```diff
@@ -33,7 +38,12 @@ def get_object(self, queryset=None):
             start_time__year=self.kwargs.get("year"),
             start_time__month=self.kwargs.get("month"),
         )
-        return queryset.get()
+
+        try:
+            obj = queryset.get()
+        except queryset.model.DoesNotExist:
+            raise Http404("No event found matching the query")
+        return obj
```
tim-schilling:
```suggestion
        return get_object_or_404(queryset)
```
We can make this a bit simpler with the shortcut function.

home/admin.py
```diff
@@ -46,6 +46,108 @@
...
+        modify, and save it as a completely new event in a single step.
+        RSVPs (rsvped_members) are intentionally excluded.
+        """
+        initial = super().get_changeform_initial_data(request)
+        copy_from = request.GET.get("copy_from")
+        if not copy_from:
+            return initial
+        try:
+            source = Event.objects.prefetch_related(
+                "speakers", "organizers", "tags"
```
tim-schilling: The `"tags"` prefetch related isn't reused since we're using the `.names()` queryset filtering method below. This needs to be adjusted to be a `Prefetch()` object, or we can pay the extra query penalty for that field.

home/admin.py
```diff
@@ -46,6 +46,108 @@
...
+
+        When the request contains a ``copy_from`` GET parameter, the add form
+        is pre-filled with the source event's data so the user can review,
+        modify, and save it as a completely new event in a single step.
+        RSVPs (rsvped_members) are intentionally excluded.
+        """
+        initial = super().get_changeform_initial_data(request)
+        copy_from = request.GET.get("copy_from")
+        if not copy_from:
+            return initial
```
tim-schilling: I think you could rework this method so that the `.update()` is protected by the various if statements and there's only one `return initial` at the end. It may be a bit easier to follow structured like that.

home/admin.py
```diff
@@ -46,6 +46,108 @@
...
+                    .distinct()
+                )
+            elif event.is_public:
+                recipients = list(
+                    User.objects.filter(profile__receiving_event_updates=True)
+                    .exclude(email="")
+                    .values_list("email", flat=True)
+                )
+            else:
+                recipients = []
```
tim-schilling: I'd like to see this logic exist on `Event` as a method.

tim-schilling (commented):
This looks like it's on the right path. I have a few small code location changes I'd recommend. I didn't look at the tests.

Either in this PR we'll also want to do the following or create follow-up issues:
- Have an admin action on SessionAdmin to generate the welcome ceremony events
- Have an admin action on SessionAdmin to generate a guest speaker event

JohananOppongAmoateng:
@tim-schilling sorry will get up to speed on this pr hopefully before the weekend need to work on some other stuff

home/models/event.py
JohananOppongAmoateng: I wasnt aware there is an arrayfield my bad. I make change it

JohananOppongAmoateng:
I will create separate prs and issues for the session admin changes

home/admin.py
```diff
@@ -46,6 +46,89 @@
...
+        initial = super().get_changeform_initial_data(request)
+        copy_from = request.GET.get("copy_from")
+        if copy_from:
+            try:
+                source = Event.objects.prefetch_related("speakers", "organizers").get(
+                    pk=copy_from
+                )
+            except Event.DoesNotExist:
+                pass
+            else:
```
tim-schilling:
```suggestion
            source = get_object_or_404(Event.objects.prefetch_related("speakers", "organizers"), pk=copy_from)
```

home/admin.py
```diff
@@ -46,6 +46,89 @@
...
+        - Session event: all members of that session who have an email address.
+        - Public event (no session): all users opted in to event updates.
+        - Private event (no session): no extra recipients; the task still sends
+          to the event's ``extra_emails`` (e.g. sessions@djangonaut.space).
+        """
+        queued = 0
+        for event in queryset:
+            tasks.send_event_calendar_invite.enqueue(
+                event_id=event.pk,
+                recipients=event.get_calendar_invite_recipients(),
```
tim-schilling: Could this be determined/called in `send_event_calendar_invite()` instead?

home/tasks/create_zoom_meeting.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    if not zoom_enabled():
+        logger.warning(
+            "Zoom credentials not configured; skipping meeting creation for event %s",
+            event_id,
+        )
+        return
+
+    event = Event.objects.filter(pk=event_id).first()
```
tim-schilling: We should do this in a [transaction using `select_for_update()`](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#select-for-update). Let me know if you'd like some help setting it up.

home/tasks/create_zoom_meeting.py
```diff
@@ -0,0 +1,67 @@
...
+    - deleted events
+    - concurrent updates
+    """
+
+    if not zoom_enabled():
+        logger.warning(
+            "Zoom credentials not configured; skipping meeting creation for event %s",
+            event_id,
+        )
+        return
```
tim-schilling, line 39: I think it may be helpful to support retriggering this task from the admin. For example, what do we do if any of these early returns are used? How does the user get the system out of that error state?

home/tasks/event_notifications.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django_tasks import task
+
+from home import email
+from home.icalendar_utils import generate_icalendar
+from home.models import Event
+
+
+@task()
+def send_event_calendar_invite(event_id: int, recipients) -> None:
```
tim-schilling: Sorry to throw another thing at you, but how do we make sure this task is idempotent, meaning how do we make sure running it N times ends up in the same state. Right now, it's possible for this task to run multiple times and email people several times. I think we would want this to send the invites out once.

home/tasks/event_notifications.py
```diff
@@ -0,0 +1,37 @@
...
+        "event": event,
+        "cta_link": event.get_full_url(),
+    }
+    recipients = recipients + list(event.extra_emails or [])
+    email.send(
+        email_template="event_calendar_invite",
+        recipient_list=recipients,
+        context=context,
+        attachments=[("event.ics", ical_data, "text/calendar")],
+    )
```
tim-schilling, line 45: This is going to show everyone's email to everyone which we probably don't want. This is because `recipient_list` is used on the to field for the email. Do you have ideas on how to avoid that?

home/templates/email/event_calendar_invite/body.txt
```diff
@@ -0,0 +1,20 @@
+You're invited: {{ event.title }}
+
+Hello,
+
+You are invited to join us for {{ event.title }} on {{ event.start_time|date:"l, F j, Y" }} at {{ event.start_time|date:"g:i A e" }}.
```
tim-schilling: Any place we show dates, let's include a link to the time.is page. We have the `time_is_link` template tag to make this easier.

home/views/events.py
```diff
@@ -22,9 +23,13 @@ class EventDetailView(DetailView):
...
     template_name = "home/event_detail.html"
 
+    def get_queryset(self):
+        """Restrict queryset to events visible to the current user."""
+        return Event.objects.for_user(self.request.user)
+
     def get_object(self, queryset=None):
         """Get event by slug, year, and month."""
-        if not queryset:
+        if queryset is None:
```
tim-schilling, line 32: Nice catch.

indymeet/settings/base.py
```diff
@@ -284,3 +284,14 @@ class FormRenderer(TemplatesSetting):
...
+# Zoom Server-to-Server OAuth credentials for automatic meeting creation.
+# Set all three to enable; leave unset to disable the feature.
+# Optionally set ZOOM_MEETING_TEMPLATE_ID to use a saved meeting template;
+# otherwise meetings are created with mute_upon_entry and waiting_room enabled.
+ZOOM_ACCOUNT_ID = os.environ.get("ZOOM_ACCOUNT_ID")
+ZOOM_CLIENT_ID = os.environ.get("ZOOM_CLIENT_ID")
+ZOOM_CLIENT_SECRET = os.environ.get("ZOOM_CLIENT_SECRET")
+
+# not really sure whether the template ids should be stored by us instead of in settings
+ZOOM_MEETING_TEMPLATE_ID = os.environ.get("ZOOM_MEETING_TEMPLATE_ID")
```
tim-schilling, line 297: This is a good idea. Thank you for adding it in. We'll adjust the comment here to match the title in zoom. I can handle that when we get to it.

tim-schilling:
Thanks @JohananOppongAmoateng I reviewed all the non-tests pretty thoroughly with an eye for production use. Please let me know if you'd like me to explain anything in more detail.

home/admin.py
JohananOppongAmoateng: Do you mean doing the loop in send_event_calender_invite?

home/tasks/event_notifications.py
JohananOppongAmoateng: i was thinking bcc?

tim-schilling: That would solve it!

home/admin.py
tim-schilling: I meant `send_event_calendar_invite()` would call `event.get_calendar_invite_recipients()` so you can remove `recipients` as an argument for the task. I'm not sure what is meant by "doing the loop".

JohananOppongAmoateng: I understand

home/templates/email/event_calendar_invite/body.txt
JohananOppongAmoateng: I dont think we can use tags in text? or is that possible

tim-schilling: We can as long as we're using the django template language to render it, which we are via `render_to_string()`

raffaellasuardini:
HI @JohananOppongAmoateng, I'm testing this locally, and I will share my initial thoughts.

home/email.py
raffaellasuardini:
```suggestion
		bcc_list = [
            recipient
            for recipient in (bcc_list or [])
            if recipient in settings.ALLOWED_EMAILS_FOR_TESTING
        ]
        if not recipient_list and not bcc_list:
```
this is useful for testing because no error is raised if you're trying locally

home/admin.py
```diff
@@ -46,6 +46,117 @@
 class EventAdmin(DescriptiveSearchMixin, admin.ModelAdmin):
     model = Event
     filter_horizontal = ("speakers", "rsvped_members", "organizers")
+    actions = ["copy_event", "send_calendar_invites", "retry_zoom_meeting_creation"]
+
```
raffaellasuardini, line 51:
```suggestion
    list_display = ["title", "start_time","calendar_invites_sent_at"]
```
since emails are only sent when the 'calendar_invites_sent_at' field has not yet been set, I find it useful to be able to view this field here. Otherwise, this information is only visible when editing the event.

home/tasks/create_zoom_meeting.py
tim-schilling: @JohananOppongAmoateng this comment is still applicable.

tim-schilling: Ack wrong one. You handled it here.

home/tasks/event_notifications.py
tim-schilling: @JohananOppongAmoateng this comment is still applicable. I'm thinking a `select_for_update()` would handle this similar to what you did for the event links.

home/templates/email/event_calendar_invite/body.txt
```diff
@@ -0,0 +1,21 @@
...
+Hello,
+
+You are invited to join us for {{ event.title }} on {{ event.start_time|date:"l, F j, Y" }} at ({{ event.start_time|time_is_link }}).
+
+{% if event.description %}{{ event.description }}
+
+{% endif %}Location: {{ event.location }}
+
+Start: {{ event.start_time| time_is_link }}
+End: {{ event.end_time | time_is_link }}
```
tim-schilling: I think the time_is_link usages here are being used in place of rendering the actual time. I don't think we want to go that far. Can we show both?

tim-schilling:
Looks good Johanan. I think it's only those last two things. I still need to run through this locally, but I can make any changes at that point on my own and so I don't have to bother you.

home/templates/email/event_calendar_invite/body.txt
JohananOppongAmoateng: yes sure i will revert that

tim-schilling:
@JohananOppongAmoateng I was working through some of the flows and realized we were missing some documentation and tests. I've started adding them. I need to work through a few more yet.

Though in the future, I think we need to start adding a test plan section to the PR :grin: It's taking me a while to remember what exactly we're building here.

tim-schilling (approved):
Looks good. When I put this on staging, we'll need to test that the email's .ics file works as expected.

tim-schilling:
Thank you @JohananOppongAmoateng!

PR #706, opened 2026-03-05
Allow users to reset their password with an unuseable password.

tim-schilling:
We created users with unusuable passwords to record their session participation. These people should be notified of this so they can change their password and/or delete their accounts.


tim-schilling:
@raffaellasuardini I'm going to merge this. If you spot something that needs to change, I can make it later.

PR #708, opened 2026-03-06
Auto-bump question ordering on conflict in survey admin

### LINKED ISSUE #707: Survery Question Ordering
It would be wonderful to have a better user experience for ordering questions in a survey. 

Current behavior: 
On the `django-admin/home/survey/7/change/`, to set the order of a question the user should choose an integer to represent the order that the question should appear in. 

Problem: 
If a question is added later, the user must go through and update the order of all the following questions. 

Desired behavior: 
Can we drag and drop them or find an easier way to handle this.

tim-schilling:
When a question is saved at a position already occupied in the same survey, all questions at that position and above shift up by one. This applies via both QuestionAdmin and the SurveyAdmin inline.

Closes #707

### DIFF
--- a/home/admin.py
+++ b/home/admin.py
@@ -648,6 +648,7 @@ class QuestionAdmin(DescriptiveSearchMixin, admin.ModelAdmin):
     readonly_fields = ("key",)
 
     def save_model(self, request, obj: Question, form, change: bool) -> None:
+        # Re-adjust the ordering on other questions of the survey.
         Question.bump_ordering(obj.survey_id, obj.ordering, exclude_pk=obj.pk)
         super().save_model(request, obj, form, change)
 
--- a/home/admin.py
+++ b/home/admin.py
@@ -647,11 +647,30 @@ class QuestionAdmin(DescriptiveSearchMixin, admin.ModelAdmin):
     )
     readonly_fields = ("key",)
 
+    def save_model(self, request, obj: Question, form, change: bool) -> None:
+        Question.bump_ordering(obj.survey_id, obj.ordering, exclude_pk=obj.pk)
+        super().save_model(request, obj, form, change)
+
 
 @admin.register(Survey)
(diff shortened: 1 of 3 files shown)


home/admin.py
```diff
@@ -647,11 +647,30 @@ class QuestionAdmin(DescriptiveSearchMixin, admin.ModelAdmin):
     )
     readonly_fields = ("key",)
 
+    def save_model(self, request, obj: Question, form, change: bool) -> None:
+        Question.bump_ordering(obj.survey_id, obj.ordering, exclude_pk=obj.pk)
```
tim-schilling, line 652:
```suggestion
        # Re-adjust the ordering on other questions of the survey.
        Question.bump_ordering(obj.survey_id, obj.ordering, exclude_pk=obj.pk)
```

home/tests/test_survey_admin.py
```diff
@@ -366,3 +366,149 @@ def test_import_csv_view_post_invalid(self):
...
+
+    def __init__(self, instances, deleted_objects=None):
+        self._instances = instances
+        self.deleted_objects = deleted_objects or []
+
+    def save(self, commit=True):
+        return self._instances
+
+    def save_m2m(self):
+        pass
```
tim-schilling: Gross. Try to get claude to not use a mock.

home/tests/test_survey_admin.py
```diff
@@ -366,3 +366,149 @@ def test_import_csv_view_post_invalid(self):
...
+    """Test that saving a question at a conflicting ordering bumps later questions."""
+
+    def setUp(self):
+        self.user = User.objects.create_superuser(
+            username="admin2", email="admin2@example.com", password="test"
+        )
+        self.survey = Survey.objects.create(name="Ordering Test Survey")
+        self.q1 = Question.objects.create(
+            survey=self.survey, label="Q1", ordering=1, type_field=TypeField.TEXT
+        )
```
tim-schilling: Remind claude about the factories

PR #723, opened 2026-03-21
Improve availability selector for mobile

### LINKED ISSUE #627: Improve the availability selector for mobile
The availability selector is a bit condensed on mobile and the unselect option requires a long select. Ideally we'd make it easier to select and unselect times.

~~For an intermediate step, we should update the help text to indicate unselecting requires a long click on mobile.~~

aadeina:
Fixes #627

  Cells were too small to tap on mobile and deselecting required a long press.       
                                                                                                                                                                                                                                         
  **Changes:**                                                                                                                                                 
  - Cell height: `h-4` → `h-8 sm:h-4` (taller on mobile, unchanged on desktop)
  - Tap-to-toggle via `touchstart`/`touchend` so a single tap selects or deselects                                                                             
  - Help text updated to reflect mobile behavior 
 
| Before | After |
|--------|-------|
| ![Before](https://github.com/user-attachments/assets/18c3a4df-0ca8-4df8-89a2-cf431f3679b4) | ![After](https://github.com/user-attachments/assets/0c415431-00b2-4c4e-88af-6559fe68b698) |                                     

Playwright test added for mobile tap behavior.


indymeet/templates/registration/update_availability.html
```diff
@@ -50,7 +50,7 @@ <h1 class="text-5xl mb-6">{% translate "Set Your Availability" %}</h1>
             <p class="text-sm">
                 <strong>{% translate "Instructions:" %}</strong>
                 {% blocktranslate %}
-                    Click and drag to select time slots when you're generally available each week. To remove selections, click the cell. Mobile devices require a long click. Times are shown in your local timezone<span id="local-timezone"></span>.
+                    Click and drag to select time slots when you're generally available each week. To remove selections, click the cell. On mobile, tap a cell to toggle it on or off, or swipe to select a range. Times are shown in your local timezone<span id="local-timezone"></span>.
```
tim-schilling: Do we need specific mobile instructions still?

tests/test_end_to_end.py
```diff
@@ -523,6 +523,70 @@ def test_availability_workflow(self, page: Page, authenticated_user):
         expect(monday_9am).not_to_have_class(re.compile(r".*\bselected\b.*"))
         expect(monday_10am).not_to_have_class(re.compile(r".*\bselected\b.*"))
 
+    def test_availability_mobile_tap_toggle(
```
tim-schilling, line 526: Please don't add a new test. There's a significant set up and tear down cost for each playwright test. Instead this should be wired into the existing availability tests. Take a look at the existing tests to see how to structure this. If you get lost, please let me know. I'm happy to elaborate more if needed.

tests/test_end_to_end.py
```diff
@@ -523,6 +523,70 @@ def test_availability_workflow(self, page: Page, authenticated_user):
...
+        page.wait_for_load_state("networkidle")
+
+        # Wait for the grid to be fully rendered
+        page.locator("#availability-grid tbody tr").first.wait_for(state="visible")
+        page.locator(".time-slot[data-day][data-hour]").first.wait_for(state="visible")
+
+        # Tap Monday 10:00 AM to select it
+        monday_10am = page.locator('.time-slot[data-day="1"][data-hour="10"]')
+        monday_10am.tap()
+        page.wait_for_timeout(200)
```
tim-schilling, line 560: We should avoid these general wait statements. It can make end to end tests brittle. Instead it should wait on the change expected in the UI.

tim-schilling:
Did you look into whether [Viselect](https://simonwep.github.io/viselect/pages/api-reference.html) better supports this event type out of the box? It seems weird that we'd have to handle this manually.

indymeet/templates/registration/update_availability.html
aadeina: Nope, unified them. Same instructions for all devices now.

tests/test_end_to_end.py
aadeina: Will drop them. Thanks for catching that.

tests/test_end_to_end.py
aadeina: Makes sense, thanks for the guidance. 
Quick question: Playwright needs a separate browser context for touch(has_touch=True + mobile viewport),  
so I'd need to create a new context and login again inside the same test. 
That work for you or do you handle it differently?

aadeina:
Viselect already has singleTap enabled by default so my manual touch handler was redundant. Thanks for pushing on this.

tests/test_end_to_end.py
tim-schilling: Ah, sorry. I wasn't aware of that. Yes, that should be fine.

tests/test_end_to_end.py
tim-schilling: @aadeina I noticed you requested another review (and I'm super late on providing it), but these are still around. If we need them, that's fine, but I'd like to understand why they're necessary.

Issue #736, opened 2026-04-16
Sync contacts with buttondown


tim-schilling:
We should sync the users who are subscribed to the newsletter to Buttondown. We also need to update user's tags and metadata automatically when a SessionMembership is changed for those users.

Let's add a tag in Buttondown for "Member" to identify them as part of the organization

PR #744, opened 2026-04-29
Revise the profile newsletter fields

tim-schilling:
This is the first leg of #736.

This removes the event and program update fields since we aren't using those currently. We add a `interested_in` field that will be control which tags will be used in Buttondown. 

### Test Plan

1. Test sign up works
2. Login with new credentials
3. Check the correct interested in value shows up
4. Update the profile with a different set of values, confirm detail view is updated.


tim-schilling:
Screenshot 2026-04-28 at 8 42 25 PM
Screenshot 2026-04-28 at 8 44 18 PM
Screenshot 2026-04-28 at 8 45 15 PM

accounts/forms.py
```diff
@@ -9,6 +9,13 @@
 from .models import UserAvailability
 from .models import UserProfile
 
+INTERESTED_IN_FIELDS = (
+    ("interested_in_djangonaut", UserProfile.DJANGONAUT),
+    ("interested_in_captain", UserProfile.CAPTAIN),
+    ("interested_in_navigator", UserProfile.NAVIGATOR),
+    ("interested_in_organizer", UserProfile.ORGANIZER),
```
tim-schilling: Let's extract out `UserProfile.DJANGONAUT` and `SessionMembership.DJANGONAUT` into constants

accounts/models.py
```diff
@@ -158,6 +163,16 @@ class UserProfile(models.Model):
         (ORGANIZER, "Organizer"),
     )
 
+    DJANGONAUT = "Djangonaut"
+    CAPTAIN = "Captain"
+    NAVIGATOR = "Navigator"
+    ROLE_INTERESTS = (
```
tim-schilling: I don't think this is used?

accounts/models.py
```diff
@@ -182,15 +197,20 @@ class UserProfile(models.Model):
...
-    receiving_program_updates = models.BooleanField(default=False)
     github_username = models.CharField(
         max_length=39,
         blank=True,
         null=False,
         default="",
         help_text="Your GitHub username (required for participation)",
     )
+    interested_in = ArrayField(
+        models.CharField(max_length=64),
```
tim-schilling: Add a validator here to confirm it's one of the constants.

accounts/views.py
```diff
@@ -129,11 +130,10 @@ def form_valid(self, form):
...
-        user.profile.receiving_program_updates = form.cleaned_data[
-            "receive_program_updates"
-        ]
-        user.profile.receiving_event_updates = form.cleaned_data[
-            "receive_event_updates"
+        user.profile.interested_in = [
+            role
+            for field_name, role in INTERESTED_IN_FIELDS
+            if form.cleaned_data.get(field_name)
         ]
```
tim-schilling: Push this down into the `clean()` method

accounts/views.py
```diff
@@ -218,17 +212,15 @@ def form_valid(self, form):
...
```
tim-schilling: Same as above, push down into `clean()`

PR #765, opened 2026-05-20
Sync Event saves to Zoom and Discord scheduled events

MarcusL11:
Feature covers: 

- Add worker service to docker compose: introduce a dedicated worker service so background tasks (Zoom/Discord sync, calendar invites, etc.) run alongside web, mirroring production’s async setup.

- Sync Event saves to Zoom and Discord scheduled events (+ admin polish): add Discord API client/service, create/update tasks, post-save routing, Django admin retry actions, and read-only fields/help text so operators can see sync status and recover from failures.


tim-schilling:
@MarcusL11 I think this may have been based off the main branch rather than develop. The PR has a bunch of merge commits. I think you can create a new branch from `develop` then cherry-pick the other commits from this branch, then force push to here it'll fix things. I'm handwaving things so let me know if you'd like a little clearer instructions.

home/admin.py
```diff
@@ -52,8 +52,55 @@
...
+    ]
+    list_display = [
+        "title",
+        "start_time",
+        "calendar_invites_sent_at",
+        "zoom_synced_at",
+        "discord_synced_at",
+    ]
+    readonly_fields = ("zoom_synced_at", "discord_synced_at")
+    fields = (
```
tim-schilling: Let's try removing `fields` to avoid new model fields from not being exposed in the admin in the future. I think we can live without `readonly_fields` if necessary. It's worth trying to allow all new fields to `Event` to automatically be exposed in the admin.

home/admin.py
```diff
@@ -158,6 +205,32 @@ def retry_zoom_meeting_creation(self, request, queryset) -> None:
...
 
+    @admin.action(description="Retry Discord event creation")
+    def retry_discord_event_creation(self, request, queryset) -> None:
+        """Retry creating a Discord event for events that don't have one."""
+        queued = 0
+        for event in queryset:
+            if event.discord_event_id:
+                continue
+            if not event.zoom_link:
+                continue
```
tim-schilling: Message user here that this event needs a zoom link, include the action necessary for that in the message ideally.

home/admin.py
```diff
@@ -158,6 +205,32 @@ def retry_zoom_meeting_creation(self, request, queryset) -> None:
                 messages.WARNING,
             )
 
+    @admin.action(description="Retry Discord event creation")
+    def retry_discord_event_creation(self, request, queryset) -> None:
+        """Retry creating a Discord event for events that don't have one."""
+        queued = 0
+        for event in queryset:
+            if event.discord_event_id:
+                continue
```
tim-schilling: Message user here that it's already created.

home/admin.py
```diff
@@ -158,6 +205,32 @@ def retry_zoom_meeting_creation(self, request, queryset) -> None:
...
+                f"Discord event creation queued for {queued} event(s).",
+                messages.SUCCESS,
+            )
+        else:
+            self.message_user(
+                request,
+                "Selected events either already have a Discord event or are "
+                "missing a Zoom link.",
+                messages.WARNING,
+            )
```
tim-schilling: This message can be removed since we'll cover this case above now.

home/integrations/discord/client.py
```diff
@@ -0,0 +1,92 @@
...
+                "Discord API request failed: %s %s -> %s: %s",
+                method,
+                url,
+                exc.response.status_code if exc.response is not None else "?",
+                body,
+            )
+            raise
+        except requests.RequestException:
+            logger.exception("Discord API request failed: %s %s", method, url)
+            raise
```
tim-schilling: Might be worth removing? Not sure. It's not hurting a ton, but seems unnecessary for our use case.

tim-schilling (commented):
Thank you @MarcusL11, I'm still working through this, but leaving these comments for now.

home/integrations/discord/client.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def create_scheduled_event(
```
tim-schilling, line 53: These methods should have a docstring that explains what they are doing. Typically it'll also provide information about what's happening in the integrated service.

home/integrations/zoom/client.py
```diff
@@ -158,3 +158,26 @@ def create_meeting(
...
+
+    def patch_meeting(
+        self,
+        *,
+        meeting_id: str,
+        topic: str,
+        start_time: datetime,
+        duration_minutes: int,
+    ) -> None:
+        """PATCH /meetings/{id}. Zoom returns 204 on success."""
```
tim-schilling: It's probably worth mentioning that this is for updating meetings with only the fields specified.

home/integrations/discord/service.py
```diff
@@ -0,0 +1,45 @@
...
+# RE: https://docs.discord.com/developers/resources/guild-scheduled-event
+def update_event(event) -> None:
+    """PATCH the Discord scheduled event with the current Event fields."""
+    discord_client.modify_scheduled_event(
+        guild_id=settings.DISCORD_GUILD_ID,
+        event_id=event.discord_event_id,
+        payload={
+            "name": event.title[:100],
+            "description": (event.description or "")[:1000],
+            "entity_metadata": {"location": event.zoom_link[:100]},
```
tim-schilling: These shouldn't silently truncate text, there should be some type of message/alert/error that occurs when it happens. The title and description are pretty safe, but the zoom link would be very problematic if it was broken.

home/integrations/discord/service.py
```diff
@@ -0,0 +1,45 @@
...
+        start_time=event.start_time,
+        end_time=event.start_time + EVENT_DURATION,
+    )
+    return str(data["id"])
+
+
+# discord enforces strict field-length caps on scheduled events.
+# RE: https://docs.discord.com/developers/resources/guild-scheduled-event
+def update_event(event) -> None:
+    """PATCH the Discord scheduled event with the current Event fields."""
```
tim-schilling: Not every developer is going to understand what "PATCH" means. It's also not clear that this is even using a rest-ful API. It's probably more useful to use less programming specific terminology in this doc string.

home/integrations/zoom/service.py
```diff
@@ -26,4 +26,4 @@ def create_event_meeting(event) -> str:
         duration_minutes=duration_minutes,
     )
 
-    return meeting["join_url"]
+    return meeting["join_url"], str(meeting["id"])
```
tim-schilling: If we need more than just one field from the meeting, maybe we should return a custom type? I typically reach for dataclasses, but I heard `NamedTuple` works a bit better in these cases because it's slimmer, while still covering the functionality needed.

home/receivers.py
```diff
@@ -1,20 +1,30 @@
...
 ) -> None:
-    """
-    Queue Zoom meeting creation when a new event is created.
+    """Dispatch Zoom and Discord side effects on Event save.
 
-    Only triggers if the event does not already have a Zoom link.
+    Create: enqueue Zoom create (which self-chains Discord create after success);
+            if admin pre-filled zoom_link, skip Zoom and enqueue Discord directly.
+    Update: enqueue update tasks for any external system the event is linked to.
+            PATCH calls are idempotent, so we don't need to detect which fields changed.
```
tim-schilling:
```suggestion
```
This isn't clear what it's referring to unless you understand the functionality within the clients. I'm not sure it provides a benefit if we have to go a few levels deep in the code to understand the comment.

home/receivers.py
```diff
@@ -1,20 +1,30 @@
...
+@receiver(post_save, sender=Event, dispatch_uid="home.sync_event_external")
+def sync_event_external_systems(
     sender, instance: Event, created: bool, **kwargs
 ) -> None:
-    """
-    Queue Zoom meeting creation when a new event is created.
+    """Dispatch Zoom and Discord side effects on Event save.
 
-    Only triggers if the event does not already have a Zoom link.
+    Create: enqueue Zoom create (which self-chains Discord create after success);
```
tim-schilling: By self-chains, do you mean

1. In admin, user creates event
2. This handler fires, schedules task to create zoom, but not discord
3. This handler fires again in the create zoom task, scheduling task to create discord

Does this just continually loop? When does it determine it doesn't need to make a save call? Can we make sure we have a test that confirms this doesn't continually loop?

home/tasks/create_discord_event.py
```diff
@@ -0,0 +1,57 @@
...
+        try:
+            discord_id = create_event(event)
+        except Exception:
+            logger.exception("Failed to create Discord event for event %s", event_id)
+            return
+
+        Event.objects.filter(pk=event_id).update(
+            discord_event_id=discord_id,
+            discord_synced_at=timezone.now(),
+        )
```
tim-schilling: Does this work properly? I would have expected the `select_for_update()` to cause problems with this.

I think the structure of this is a bit of a footgun where we expect to avoid the signal to be raised before triggering the event. It may be better for us to create a specific factory that's responsible for creating/updating an event that will handle the downstream events. I'm not suggesting we make this change right now, I'd like to hear your thoughts on it first.

home/tests/test_discord.py
```diff
@@ -0,0 +1,404 @@
...
+
+        self.assertEqual(payload["name"], "My Event")
+        self.assertEqual(payload["description"], "A description")
+        self.assertEqual(payload["entity_type"], 3)
+        self.assertEqual(payload["privacy_level"], 2)
+        self.assertEqual(
+            payload["entity_metadata"], {"location": "https://zoom.us/j/123"}
+        )
+        self.assertEqual(payload["scheduled_start_time"], "2026-06-01T14:00:00+00:00")
+        self.assertEqual(payload["scheduled_end_time"], "2026-06-01T15:00:00+00:00")
```
tim-schilling: Is there a reason this isn't a single `assertEqual` with a dictionary?

home/tests/test_discord.py
```diff
@@ -0,0 +1,404 @@
...
+
+        mock_discord.enqueue.assert_called_once_with(event_id=event.pk)
+
+
+class DiscordServiceConstantsTests(TestCase):
+    def test_event_duration_is_one_hour(self):
+        from datetime import timedelta
+
+        self.assertEqual(EVENT_DURATION, timedelta(hours=1))
+
```
tim-schilling:
```suggestion
```

This one isn't necessary.

home/tests/test_discord.py
```diff
@@ -0,0 +1,404 @@
...
+        mock_update.side_effect = Exception("Discord down")
+
+        update_discord_event.call(event_id=self.event.pk)
+
+        self.event.refresh_from_db()
+        self.assertIsNone(self.event.discord_synced_at)
+
+
+class UpdateReceiverTests(TestCase):
+    """Edit-time receiver behavior: enqueue update tasks for linked systems."""
```
tim-schilling: I'm not sure what "edit-time receiver behavior" means.

How are these tests different than those in `DiscordSignalTests`? Can we combine them? Are there duplicates that can be reduced?

home/tests/test_zoom.py
```diff
@@ -368,3 +370,107 @@ def test_updating_existing_event_does_not_create_new_meeting(self, mock_create):
...
+        mock_patch.side_effect = Exception("Zoom down")
+
+        update_zoom_meeting.call(event_id=self.event.pk)
+
+        self.event.refresh_from_db()
+        self.assertIsNone(self.event.zoom_synced_at)
+
+
+class Phase4RecursionGuardTests(TestCase):
+    """Phase 2's recursion-guard test, deferred until update tasks exist."""
```
tim-schilling: I don't understand the name or docstring here around "phase"

home/tests/test_zoom.py
```diff
@@ -368,3 +370,107 @@ def test_updating_existing_event_does_not_create_new_meeting(self, mock_create):
...
+
+    @override_settings(**ZOOM_SETTINGS)
+    @patch("home.tasks.create_zoom_meeting.create_discord_event")
+    @patch("home.tasks.create_zoom_meeting.create_event_meeting")
+    @patch("home.tasks.update_discord_event")
+    @patch("home.tasks.update_zoom_meeting")
+    def test_zoom_writeback_does_not_recurse(
```
tim-schilling: This is probably too much mocking to be useful as an integration test, which is what I suspect this is meant to be.

tim-schilling (changes requested):
Thanks for pushing this forward. I think most of my concerns stem around patterns originally implemented in the zoom version that are being propagated and somewhat extended. I don't think we should continue those without more consideration first.

home/integrations/discord/client.py
MarcusL11: Got it, its redundant -- Is it because [create_discord_event](https://github.com/djangonaut-space/wagtail-indymeet/pull/765/changes#diff-4aa7c16d9fbd80397564b13cd289a12eb9e3bd391df5c62b78bf0e925932e6e6R48) caller catches the Exception already

home/integrations/discord/service.py
tim-schilling: As per our conversation, we landed on moving the name and description limitations to the Event model itself. If we discover in the future that we need to have longer names and descriptions for things outside of discord, we can adjust it then.

Edit:
The zoom link will truncate, but it will use `logger.error()` to send a Sentry email to maintainers. It's very unlikely that it will happen since Zoom renders links that are ~72 characters long.

home/integrations/zoom/service.py
tim-schilling: As per our meeting, we'll create this class in this file to keep it closely located to where it gets used.

home/tasks/create_discord_event.py
tim-schilling: As per our meeting, we're going to have a single task that's responsible for creating/updating the zoom link, then creating/updating the discord event. This task will be triggered by an operations or factories function that will be used in `EventAdmin.save_model()`. This will make it very clear how these services get triggered downstream without using signals or chained tasks (which are impossible with the current implementation of tasks).

home/receivers.py
tim-schilling: Resolving. See https://github.com/djangonaut-space/wagtail-indymeet/pull/765#discussion_r3321323840 for why.

home/tests/test_zoom.py
tim-schilling: This will likely disappear with the changes in https://github.com/djangonaut-space/wagtail-indymeet/pull/765#discussion_r3321323840

home/admin.py
```diff
@@ -100,6 +82,21 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
             formfield.help_text = self.field_help_text[db_field.name]
         return formfield
 
+    def save_model(self, request, obj, form, change) -> None:
```
tim-schilling, line 84: I like how this turned out. Nice!

home/tests/test_operations.py
```diff
@@ -0,0 +1,160 @@
...
+        event = EventFactory.create(zoom_link="")
+
+        with self.assertLogs("home.operations", level="WARNING"):
+            decision = dispatch_event_sync(event)
+
+        self.assertEqual(decision.status, EventSyncStatus.SKIPPED_NO_ZOOM_CONFIGURED)
+        mock_task.enqueue.assert_not_called()
+
+
+class EventAdminSaveModelTests(TestCase):
```
tim-schilling: Let's move EventAdminSaveModelTests into test_admin_event.py

home/tests/test_operations.py
```diff
@@ -0,0 +1,160 @@
...
+
+        self.admin.save_model(request, event, form=None, change=False)
+
+        stored = list(request._messages)
+        self.assertEqual(len(stored), 1)
+        self.assertEqual(stored[0].level, messages.WARNING)
+        self.assertEqual(str(stored[0]), "No Zoom configured.")
+
+
+class EventModelValidationTests(TestCase):
```
tim-schilling: Let's move this into a file for testing models if there isn't one that exists already.

home/admin.py
```diff
@@ -223,7 +220,7 @@ def retry_discord_event_creation(self, request, queryset) -> None:
                     messages.WARNING,
                 )
                 continue
-            tasks.create_discord_event.enqueue(event_id=event.pk)
+            tasks.sync_event.enqueue(event_id=event.pk)
```
tim-schilling: I think we can consolidate the actions for retrying zoom and discord event creation into one to make things simpler for the user.

tim-schilling:
The last changes are around moving things around. Everything is looking good. Thank you for sticking with it while we sorted out how to best handle this.

Let me know if you'd rather have me do the remaining tweaks.

Side note: I merged a change [into develop] that switched the project to be a docker-first project. It also consolidated all the settings into one file, which is where the merge conflict is coming from.

home/tests/test_operations.py
MarcusL11: Done — moved EventAdminSaveModelTests into test_admin_event.py alongside the other EventAdmin action tests.

home/tests/test_operations.py
MarcusL11: Done. test_models.py already existed, so EventModelValidationTests now lives there

home/admin.py
MarcusL11: Done. Consolidated into a single "Retry Zoom/Discord sync" action.

tim-schilling (approved):
Thank you for making those changes!

tim-schilling:
I tweaked the commit message on the first commit to lead with the "Sync Event saves to Zoom and Discord scheduled events" message.

PR #800, opened 2026-07-02
Purge cloudflare cache on deployment

### LINKED ISSUE #785: Purge cloudflare cache on deployment
There's an integration with wagtail to do some purging with cloudflare, but it's not clearing the styles we build on deployment. We should modify our release script to purge cloudflare cache after a release goes live.

Mr-Sunglasses:
closes: #785


home/management/commands/purge_cloudflare_cache.py
```diff
@@ -0,0 +1,63 @@
...
+            self.stderr.write(
+                self.style.ERROR(f"Failed to purge Cloudflare cache: {exc}")
+            )
+            return
+
+        if not response.ok or not response_json.get("success"):
+            self.stderr.write(
+                self.style.ERROR(f"Failed to purge Cloudflare cache: {response_json}")
+            )
+            return
```
tim-schilling: This should return a 1 (I think that's the error code), so that the release process fails.

home/management/commands/purge_cloudflare_cache.py
```diff
@@ -0,0 +1,63 @@
...
+                timeout=30,
+            )
+            response_json = response.json()
+        except (requests.exceptions.RequestException, ValueError) as exc:
+            # A failed purge shouldn't fail the deployment - stale cache is
+            # recoverable, an aborted release is not.
+            self.stderr.write(
+                self.style.ERROR(f"Failed to purge Cloudflare cache: {exc}")
+            )
+            return
```
tim-schilling: Could we also use `logger.error(exc_info=True)` so that the errors go into Sentry as well? This looks like it would be logged in the deploy logs only and only the stacktrace would be included. Sentry would provide a little more visibility into things.

tim-schilling (changes requested):
Thanks for getting this written up! Could you add a couple tests so that this code has some coverage please?

Mr-Sunglasses:
> Thanks for getting this written up! Could you add a couple tests so that this code has some coverage please?



> Thanks for getting this written up! Could you add a couple tests so that this code has some coverage please?

Thanks a lot @tim-schilling for the review, I've done the changes that you've suggested and added the tests 😊

tim-schilling (approved):
Thank you @Mr-Sunglasses! I had two nitpicky things I wanted to see, but didn't feel like you needed to do them. I prefer the test mocking to not use `mock` when possible and instead to mock the HTTP requests and responses. I think this tests more of the code in the application and requires a developer to think more about the API interface. Again, nitpicky. The other bit was the test location. We don't have great organization, but I figured consistency is best.

Mr-Sunglasses:
> Thank you @Mr-Sunglasses! I had two nitpicky things I wanted to see, but didn't feel like you needed to do them. I prefer the test mocking to not use `mock` when possible and instead to mock the HTTP requests and responses. I think this tests more of the code in the application and requires a developer to think more about the API interface. Again, nitpicky. The other bit was the test location. We don't have great organization, but I figured consistency is best.

Thanks, @tim-schilling! 😊 I'll make sure to incorporate your suggestions in my next PR

PR #802, opened 2026-07-04
fix: cross-env: not found error in docker-dev setup

### LINKED ISSUE #801: bug:  `cross-env: not found` error in docker-compose
While running the dev setup via docker-compose, it gives the error `cross-env: not found`.

I dug a little into this bug and found that we installs npm dependencies into the image at `/app/theme/static_src/node_modules`, but the compose file bind-mounts the host repo over `/app`. Since the host has no 	`node_modules`, the bind mount hid the installed packages, so `npm run dev` couldn't find `cross-env`.

Mr-Sunglasses:
closes: #801


tim-schilling:
Thank you @Mr-Sunglasses

PR #803, opened 2026-07-04
fix: about sub-menu collapse while using tab

Mr-Sunglasses:
Before:

https://github.com/user-attachments/assets/1370fd03-9f1a-4b49-a1c4-f65a42a25d22

After:

https://github.com/user-attachments/assets/b5c2ca91-29ab-4001-bc05-db1886dccc71


tim-schilling:
@Mr-Sunglasses I'm sorry, but I'm not sure what's being changed. Those videos look the same to me. Can you describe what the bug is?

And it would be beneficial for us to include this in a playwright test to make sure this is doing what we expect it to and it doesn't break.

PR #815, opened 2026-07-14
test: add test coverage for Discord scheduled event update path

mgstabrani:
Cover the modify/update flow that mirrors the existing create tests:

- DiscordClient.modify_scheduled_event: PATCH to the scoped guild/scheduled-event URL, payload forwarded unchanged, and the updated event JSON returned.
- DiscordClient._request: logs the response body and re-raises on HTTP error.
- service.update_event: builds the same payload as the create path but as a PATCH targeting the stored discord_event_id, and raises without calling Discord when the Zoom link exceeds the location limit.

### DIFF
--- a/home/tests/test_discord.py
+++ b/home/tests/test_discord.py
@@ -228,9 +228,7 @@ class DiscordClientRequestTests(TestCase):
     def test_logs_response_body_and_reraises_on_http_error(self):
         rsps.add(rsps.GET, f"{BASE_URL}/foo", status=403, body="boom")
 
-        with self.assertLogs(
-            "home.integrations.discord.client", level="ERROR"
-        ) as logs:
+        with self.assertLogs("home.integrations.discord.client", level="ERROR") as logs:
             with self.assertRaises(requests.HTTPError):
                 self.client._request("GET", "/foo")
 
--- a/home/tests/test_discord.py
+++ b/home/tests/test_discord.py
@@ -7,7 +7,7 @@ Tests for Discord integration: DiscordClient, the discord service
 import json
 from datetime import datetime as dt
 from datetime import timezone as dt_timezone
-from unittest.mock import MagicMock, patch
+from unittest.mock import patch
 
 import requests
 import responses as rsps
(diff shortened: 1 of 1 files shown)


home/tests/test_discord.py
```diff
@@ -130,6 +132,61 @@ def test_handles_null_description(self):
         self.assertEqual(mock_req.call_args.kwargs["json"]["description"], "")
 
 
+class DiscordClientModifyScheduledEventTests(TestCase):
```
tim-schilling: I think these tests can be merged into the other test class.

home/tests/test_discord.py
```diff
@@ -157,6 +214,21 @@ def test_logs_on_429(self):
                 with self.assertRaises(Exception):
                     self.client._request("GET", "/foo")
 
+    @override_settings(**DISCORD_SETTINGS)
+    def test_logs_response_body_and_reraises_on_http_error(self):
+        resp = MagicMock(status_code=500)
+        resp.text = "boom"
+        resp.raise_for_status.side_effect = requests.HTTPError(response=resp)
```
tim-schilling: Ideally we'd redo all these tests to use the `responses` library instead of `unittest.mock`

home/tests/test_discord.py
```diff
@@ -157,6 +214,21 @@ def test_logs_on_429(self):
...
+
+        with patch.object(self.client.session, "request", return_value=resp):
+            with self.assertLogs(
+                "home.integrations.discord.client", level="ERROR"
```
tim-schilling: TIL. That's cool

home/tests/test_discord.py
```diff
@@ -187,6 +259,51 @@ def test_raises_on_over_long_zoom_link(self):
             _prepare_fields(event)
 
 
+class UpdateEventTests(TestCase):
+    """Service-level update_event: the payload sent to Discord mirrors the
+    create path, but as a PATCH targeting the stored discord_event_id."""
+
+    @override_settings(**DISCORD_SETTINGS)
+    @patch("home.integrations.discord.service.discord_client")
```
tim-schilling: Ideally, we should be mocking this at the HTTP API level.

home/tests/test_discord.py
mgstabrani: merged into `DiscordClientScheduledEventTests` along with `DiscordClientCreateScheduledEventTests`

home/tests/test_discord.py
mgstabrani: Done, converted the DiscordClient tests to stub HTTP with responses. Kept the SyncEvent* tests on unittest.mock since they mock the service functions to test task orchestration, not the HTTP layer.

home/tests/test_discord.py
mgstabrani: Fixed, update_event now runs against a stubbed HTTP endpoint via `responses`

home/tests/test_discord.py
```diff
@@ -92,70 +98,141 @@ def test_payload_shape(self):
...
+
+        self.assertEqual(rsps.calls[0].request.method, "PATCH")
+        self.assertEqual(
+            rsps.calls[0].request.url,
+            f"{BASE_URL}/guilds/123/scheduled-events/event-1",
+        )
+
+    @override_settings(**DISCORD_SETTINGS)
+    @rsps.activate
+    def test_forwards_payload_unchanged(self):
```
tim-schilling: What is this testing?

tim-schilling: Ignore me, I'll adjust it on merge.

PR #816, opened 2026-07-14
test: cover accounts admin CSV export and custom actions

mgstabrani:
Add tests for the admin behaviour that was previously uncovered:

- ExportCsvMixin.export_as_csv: returns a CSV attachment with the header and selected rows, strips sensitive/binary fields, and writes a header-only file for an empty queryset.
- CustomUserAdmin.compare_availability_action: redirects to the compare page with the selected user IDs, and messages the user (returning None) when nothing is selected.
- CustomGroupAdmin.recreate_session_organizers_group: invokes the setup_session_organizers_group management command.

### DIFF
--- a/accounts/tests/test_admin.py
+++ b/accounts/tests/test_admin.py
@@ -1,4 +1,6 @@
 from home import constants
+import csv
+import io
 from datetime import timedelta
 from http import HTTPStatus
 from unittest.mock import MagicMock, patch
@@ -154,7 +156,8 @@ class ExportCsvActionTests(TestCase):
     def setUp(self) -> None:
         self.client.force_login(self.superuser)
 
-    def test_export_returns_csv_attachment_with_selected_rows(self) -> None:
+    def test_export_csv(self) -> None:
+        """Selected rows come back as a CSV attachment with password excluded."""
         url = reverse("admin:accounts_customuser_changelist")
 
         response = self.client.post(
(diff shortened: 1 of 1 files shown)


accounts/tests/test_admin.py
```diff
@@ -137,3 +140,117 @@ def test_useravailability_updated_at_filter(self) -> None:
...
+        )
+
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+        self.assertEqual(response["Content-Type"], "text/csv")
+        self.assertIn("attachment; filename=", response["Content-Disposition"])
+
+        content = response.content.decode()
+        self.assertIn("username", content)
+        self.assertIn("admin@example.com", content)
+        self.assertNotIn("password", content)
```
tim-schilling, line 175: This lacks a testing for a core piece of that reuseable logic in the action. It's just testing a few edges around what it's doing and confirming that the thing generally runs.

tim-schilling (commented):
I'm more inclined to close this than merge it. Some of the tests it adds aren't testing real logic if that makes sense, such as `test_recreate_session_organizers_group_runs_command`. It makes the coverage % number go up, but I don't know if it makes me feel like the application is better tested. Do you know what I mean?

accounts/tests/test_admin.py
mgstabrani: Fair point. The reusable piece is ExportCsvMixin.export_as_csv and I only tested edges around it. I'll parse the CSV and assert the real field→value mapping, confirm all ignored fields (password/token/bio_image) are stripped, and cover it on UserProfileAdmin too so the "reusable across models" part is actually exercised.

mgstabrani:
> I'm more inclined to close this than merge it. Some of the tests it adds aren't testing real logic if that makes sense, such as `test_recreate_session_organizers_group_runs_command`. It makes the coverage % number go up, but I don't know if it makes me feel like the application is better tested. Do you know what I mean?

I know exactly what you mean. test_recreate_session_organizers_group_runs_command mocks call_command, so it only asserts the action calls the command — no real behavior, and that behavior is already covered in test_setup_session_organizers_group.py. I'll drop it. Let me push the reworked tests rather than leave it as-is — if it still doesn't earn its place, happy to close.

accounts/tests/test_admin.py
tim-schilling: @mgstabrani can you please communicate with me without using an LLM to generate the majority of the text. It's a bit disheartening [(I go into more detail here)](https://www.better-simple.com/django/2026/03/16/give-django-your-time-and-money/).

tim-schilling: Will fix on my end.

mgstabrani: @tim-schilling My point was that I agree with your comment that my previous code just tested whether it was running or not. Not whether it is correct or not. Thanks anyway.

Also, thanks for your reminder when using the LLM. Really appreciate it. I do agree with your statement that LLMs should be used as `a complementary tool, not as a vehicle`. When I am not sure or don't understand some stuff, I do kind of reverse things. I ask the LLM and understand through its explanation. Anyway. I will try to use LLM more wisely in future contributions.

PR #820, opened 2026-07-29
Fix issue with Update Availability page using standard timezone

tim-schilling:
Don't use a fixed time when storing the availabilities. This avoids showing the selected availabilities in standard time while the user is in DST.

When this is shipped, the following code should be used to update the availabilities so they that the users don't need to update things on their own. This isn't ideal, but should help a bit until we start storing the timezone explicitly.

```python
from datetime import timedelta
from django.utils import timezone  # noqa: E402
from accounts.models import UserAvailability  # noqa: E402
from home.availability import HOURS_PER_WEEK  # noqa: E402

LOOKBACK = timedelta(weeks=5)

def shift_back_one_hour(slot: float) -> float:
    """Shift a slot (hours from start of week, UTC) back by one hour, wrapping around the week."""
    return (slot - 1) % HOURS_PER_WEEK

def fix() -> None:
    cutoff = timezone.now() - LOOKBACK
    availabilities = (
        UserAvailability.objects.filter(updated_at__gte=cutoff)
        .exclude(slots=[])
        .select_related("user")
    )

    for availability in availabilities:
        availability.slots = sorted(
            shift_back_one_hour(slot) for slot in availability.slots
        )
        availability.updated_at = timezone.now()
        availability.save(update_fields=["slots", "updated_at"])
        print(availability.user)

```


tim-schilling:
The script has been run.

PR #825, opened 2026-08-01
Feat/support dst

MarcusL11:
This PR enhances availability handling so recurring weekly availability works correctly across DST. It changes stored availability from fixed UTC weekly slots to local wall-clock weekly slots plus the selected IANA timezone. When availability is compared across users, each user’s local slot is converted to a UTC reference slot for the relevant week, using that user’s timezone rules. The comparison grid then displays those UTC-normalized slots in the viewer’s timezone.


accounts/tests/test_availability.py
```diff
@@ -198,12 +225,60 @@ def test_form_saves_slots(self, user):
...
 
         assert form.is_valid()
         saved_availability = form.save()
 
         assert saved_availability.slots == test_slots
+        assert saved_availability.slots_timezone == CENTRAL_EUROPEAN_TIMEZONE
+
+    def test_form_timezone_choices_include_iana_zones(self, user):
+        """Test that the timezone selector offers IANA timezone names."""
+        from accounts.forms import UserAvailabilityForm
```
mgstabrani: Any reason why this is imported in function-level scope instead of global scope?

accounts/forms.py
```diff
@@ -195,18 +199,40 @@ class UserAvailabilityForm(forms.ModelForm):
...
+        fields = ("slots", "slots_timezone")
+
+    def clean_slots_timezone(self) -> str:
+        """Default omitted timezone submissions without overwriting saved rows."""
+        slots_timezone = self.cleaned_data.get("slots_timezone")
+        if slots_timezone:
+            return slots_timezone
+        if self.instance and self.instance.pk:
+            return self.instance.slots_timezone or "UTC"
+        return "UTC"
```
mgstabrani: Can we extract this `UTC` magic string into a constant?

home/views/team_formation.py
```diff
@@ -261,7 +262,9 @@ def get_filtered_applicants(
             availability_slots = user.availability.slots or []
             if availability_slots:
                 has_availability = True
-                availability_by_day = format_availability_by_day(availability_slots)
+                availability_by_day = format_availability_by_day(
+                    get_user_utc_slots(user), "UTC"
```
mgstabrani: Same as the above comment. Please extract this magic string into a constant if possible.

accounts/forms.py
tim-schilling: We should be able to use `datetime.UTC` which is an alias of `datetime.timezone.utc`

home/views/compare_availability.py
```diff
@@ -142,7 +156,7 @@ class CompareAvailabilityForm(forms.Form):
         required=False,
         widget=TomSelectMultipleWidget(),
     )
-    offset = forms.FloatField(required=False, initial=0)
+    timezone = forms.CharField(required=False, initial="UTC")
```
tim-schilling: Question: Would it help to fetch from the user?

home/availability.py
```diff
@@ -287,10 +380,11 @@ def calculate_team_overlap(
         navigator_users: List of navigator CustomUser instances
         captain_user: Captain CustomUser instance (can be None)
         djangonaut_users: List of djangonaut CustomUser instances
+        reference_date: Date inside the reference week; defaults to today
```
tim-schilling: What is this date used for exactly? I think a little more context would help a future reader a lot.

home/managers.py
```diff
@@ -392,46 +392,87 @@ def without_team_assignment(self, session):
...
+        seen_user_ids = set()
+        for response in self.select_related("user").prefetch_related(
+            "user__availability"
+        ):
+            user = response.user
+            if user.id in seen_user_ids:
+                continue
+            seen_user_ids.add(user.id)
+            if target_slots.intersection(get_user_utc_slots(user, reference_date)):
+                matching_user_ids.append(user.id)
```
tim-schilling: I think we may have problems with this on a large scale (100 users). We should look at it closely.

indymeet/templates/registration/update_availability.html
```diff
@@ -116,77 +129,56 @@ <h1 class="text-5xl mb-6">{% translate "Set Your Availability" %}</h1>
 </main>
 
 <script>
-    document.getElementById("local-timezone").textContent = ` (${Intl.DateTimeFormat().resolvedOptions().timeZone})`
+    const browserTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
```
tim-schilling: Is it possible this doesn't get set?

tim-schilling (commented):
Looked through it for a first pass. I still need to pull this down and play with it and see how it all works. I love the javascript reduction. That's nice. I suspect we want to add some unit tests around daylights savings since that's the driving concern for this.

tim-schilling:
@MarcusL11 I added a few commits to round out some of the functionality to make things easier during the transition.

accounts/forms.py
tim-schilling: Ah, we would need to switch to something like `django-timezonefield` where the attribute gets converted to a `zoneinfo` in order to use `datetime.UTC`.

home/availability.py
tim-schilling: I saw it was only used for manipulating the dates in testing. We can use the `freezegun` library for that, so I pushed a commit that removed it. Let me know if I misunderstood its purpose.

tim-schilling:
Alright, I'm done making changes for the next 12 hours 😁 Thanks for getting us here @MarcusL11. This is working really nicely so far.

tim-schilling:
I just realized that we should have [f2e2fe4](https://github.com/djangonaut-space/wagtail-indymeet/pull/825/commits/f2e2fe4fa946820aab5cfe1fedc9bcc724b76f32) automatically save rather than just adjust the values on the fly. This would get them off utc and into the correct timezone avoiding the DST problems.

accounts/tests/test_availability.py
MarcusL11: Good catch. I moved `UserAvailabilityForm` to the top level imports. I also cleaned up another local import in the same file while I was there, `IntegrityError` is now imported at module scope too.

accounts/forms.py
MarcusL11: Good call. I kept this as a string constant rather than `datetime.UTC`, since this field stores IANA timezone names as strings. I added `DEFAULT_AVAILABILITY_TIMEZONE = "UTC"` and updated the availability form comparisons to use it.

home/views/team_formation.py
MarcusL11: Updated this one too. This now uses `constants.DEFAULT_AVAILABILITY_TIMEZONE` instead of the literal `"UTC"`.

home/views/compare_availability.py
MarcusL11: I revisited this and agreed the compare view should default from the requesting user's saved availability timezone. The form now sets its default through `get_user_compare_timezone()`. If the user does not have an availability timezone, it falls back to `DEFAULT_AVAILABILITY_TIMEZONE`.

home/availability.py
MarcusL11: got it!

home/managers.py
MarcusL11: I think you're right.  So I changed it to scan UserAvailability rows directly instead of the full survey response and user objects.  The new approach builds candidate user IDs from the current qs, fetches only user_id, slots, and slots_timezone, then returns the original qs filtered by matching user IDs.

indymeet/templates/registration/update_availability.html
MarcusL11: Good point. I hardened this path so missing or unsupported `Intl` does not break the availability page. The display now wraps browser timezone detection safely, falls back to the selected timezone from the form, then finally falls back to the server provided `default_availability_timezone`. I also applied the same safe detection pattern in `availability_timezone_autodetect.js`.

tim-schilling:
I'm going to clean up the PR queue, then rebase this, clean up the conflicts, merge it, then I have a PR that refactors a `Slot` class out of everything to reduce the number of formatting and conversion methods. It should ideally make things easier to build on top of.

After that, we'll want to make the compare availability page allow changing the timezone and defaulting to the user's current timezone. Having it subtly pull what's stored for a user was odd for me.

tim-schilling:
@MarcusL11 I confirmed that the compare availability page uses the current user's timezone, even when comparing their own availability to another person and their stored availability is in another timezone. There's no bug for us to clean up.

MarcusL11:
noted! Thanks @tim-schilling

PR #866, opened 2026-08-11
Replace mutable username search with a DiscordMember guild mirror linked from UserProfile, kept fresh by hourly sync and session setup/teardown.

### LINKED ISSUE #862: Make matching Discord users to web app users easier
We still need to map users from discord to our web application. Rather than asking them to actually connect their account, we could do the following:

- Fetch all discord users on the server and store a copy in a model
- Webhook integration for any user changes (or just poll with the above)
- Switch out the discord username for this new model on `SessionMembership`
- Potentially attempt to auto-map a user when they sign up if there's an ongoing session? This seems risky.

JohananOppongAmoateng:
Replace mutable username search with a DiscordMember guild mirror linked from UserProfile, kept fresh by hourly sync and session setup/teardown.  Closes #862


JohananOppongAmoateng:
I will fix the test coverage tommorow

docs/integrations/discord.md
```diff
@@ -98,10 +98,14 @@ To get a working local setup:
...
 4. In Discord, drag the bot's role above every role it manages.
 5. Enable the **Server Members Intent** on the bot.
-6. Set Discord usernames on the profiles of any users you'll test with.
+6. Run `python manage.py sync_discord_members` (optionally with `--apply-links` to backfill from legacy `discord_username` values), then link Discord members on user profiles / session memberships in admin.
 
 Then the Session admin's setup/teardown actions run against the test server.
 
+## Member mapping
+
+Guild members are mirrored in `DiscordMember` and linked from `UserProfile.discord_member` (stable Discord snowflake, not mutable username). The mirror is refreshed hourly and again during session setup/teardown. Mapping is admin-explicit — there is no signup auto-link.
```
tim-schilling, line 107:
```suggestion
Guild members are mirrored in `DiscordMember` and linked from `UserProfile.discord_member` (stable Discord snowflake, not mutable username). The mirror is refreshed hourly and again during session setup/teardown.
```

I'm not quite sure what that text from the AI was suggesting.

home/integrations/discord/session_service.py
```diff
@@ -219,40 +218,19 @@ def _ensure_role(name: str, role_map: dict[str, str], roles_created: list[str])
...
-            ),
-            None,
-        )
-
     resolutions = {}
     for membership in memberships:
-        username = membership.user.profile.discord_username.strip()
-        guild_member = guild_members.get(username.casefold()) if username else None
+        member = membership.user.profile.discord_member
+        linked = member is not None and member.is_active and not member.is_bot
```
tim-schilling: I don't think we need to bother with these checks for what we're doing. If the profile has a `discord_member` set, that's good enough.

home/integrations/discord/session_service.py
```diff
@@ -219,40 +218,19 @@ def _ensure_role(name: str, role_map: dict[str, str], roles_created: list[str])
...
             display_name=membership.user.get_full_name() or membership.user.username,
-            discord_username=username,
-            member_id=guild_member["user"]["id"] if guild_member else None,
+            discord_username=member.username if linked else "",
+            member_id=member.discord_id if linked else None,
             role=membership.role,
-            guild_role_ids=(
-                frozenset(guild_member["roles"]) if guild_member else frozenset()
-            ),
+            guild_role_ids=frozenset(member.role_ids) if linked else frozenset(),
```
tim-schilling: We probably should just pass `member` in here rather than exposing each of the individual fields.

home/integrations/discord/session_service.py
```diff
@@ -356,9 +318,9 @@ def team_voice_channel_name(team) -> str:
 
 def _membership_mention(membership) -> str:
     """Copy/paste text for one person: ``@username``, or their name without one."""
-    username = membership.user.profile.discord_username.strip()
-    if username:
-        return f"@{username}"
+    member = membership.user.profile.discord_member
+    if member is not None and member.is_active and not member.is_bot:
+        return f"@{member.username}"
     return membership.user.get_full_name() or membership.user.username
```
tim-schilling, line 327: We can likely move this logic down into `DiscordMember` model then and drop the if statement.

home/models/discord.py
```diff
@@ -42,3 +41,29 @@ def __str__(self) -> str:
...
+    username = models.CharField(max_length=32)
+    global_name = models.CharField(max_length=64, blank=True, default="")
+    nickname = models.CharField(max_length=64, blank=True, default="")
+    role_ids = models.JSONField(default=list, blank=True)
+    is_bot = models.BooleanField(default=False)
+    is_active = models.BooleanField(
+        default=True,
+        help_text="False when the member left the server; kept so profile "
+        "links survive leave/rejoin.",
+    )
```
tim-schilling: Do we need `role_ids`, `is_bot` and `is_active`?

home/models/discord.py
```diff
@@ -42,3 +41,29 @@ def __str__(self) -> str:
...
         """The message text Discord renders as a ping for this role."""
         return f"<@&{self.discord_id}>"
+
+
+class DiscordMember(BaseModel):
+    """One member of the Discord server, keyed by stable snowflake."""
+
+    discord_id = models.CharField(max_length=32, unique=True)
+    username = models.CharField(max_length=32)
+    global_name = models.CharField(max_length=64, blank=True, default="")
```
tim-schilling: What's the difference between these two?

home/models/discord.py
```diff
@@ -42,3 +41,29 @@ def __str__(self) -> str:
...
+    nickname = models.CharField(max_length=64, blank=True, default="")
+    role_ids = models.JSONField(default=list, blank=True)
+    is_bot = models.BooleanField(default=False)
+    is_active = models.BooleanField(
+        default=True,
+        help_text="False when the member left the server; kept so profile "
+        "links survive leave/rejoin.",
+    )
+    last_seen_at = models.DateTimeField(default=timezone.now)
```
tim-schilling: As much as I would love to have this data, I don't think we need it. It will also simplify our sync process (see below).

home/services/discord_members.py
```diff
@@ -0,0 +1,116 @@
...
+    Members are matched on ``discord_id``. Leaving the server sets
+    ``is_active=False`` rather than deleting the row, so a ``UserProfile``
+    link survives leave/rejoin. Returns the raw Discord payloads so callers
+    like teardown can reuse the fetch.
+    """
+    raw_members = list_all_guild_members()
+    now = timezone.now()
+    seen_ids: list[str] = []
+    with transaction.atomic():
+        for raw in raw_members:
```
tim-schilling: I think we should optimize this such that we aren't updating a row every hour for every user since we don't have to. To do that, first pull all the users out and exclude any that don't have changes. Then for those that do, you can use `update_or_create` or if you want to reduce that down to 1 query, you could switch to [`bulk_create`](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#bulk-create) and use `update_conflicts=True`.

home/services/discord_members.py
```diff
@@ -0,0 +1,116 @@
...
+        deactivated_count = deactivated.count()
+        deactivated.update(is_active=False)
+    return MemberSyncReport(
+        synced=len(seen_ids),
+        deactivated=deactivated_count,
+        members=raw_members,
+    )
+
+
+def apply_username_links(*, dry_run: bool = False) -> LinkReport:
```
tim-schilling: I think we should scrap this idea. Given it's a public repository, the web app allows anyone to sign up and the discord server is public as well, it's certainly possible someone could use this to gain access to a privileged account.

I think if we want to "automate" this, we'd have to go the route of having the user link their discord account via an oauth login to their djangonaut.space account. We don't have to do that right now though.

Do you have any ideas on how we can make this user association easier for session organizers?

home/admin.py
```diff
@@ -576,7 +577,8 @@ class SessionMembershipAdmin(ExportMixin, DescriptiveSearchMixin, admin.ModelAdm
         "user__email",
         "user__first_name",
         "user__last_name",
-        "user__profile__discord_username",
+        "user__profile__discord_member__username",
+        "user__profile__discord_member__discord_id",
```
tim-schilling, line 579:
```suggestion
        "user__profile__discord_member__nickname",
```

home/admin.py
```diff
@@ -623,18 +625,19 @@ def github_username(self, obj: SessionMembership) -> str:
             return obj.user.profile.github_username
         return "-"
 
-    @admin.display(description="Discord", ordering="user__profile__discord_username")
-    def discord_username(self, obj: SessionMembership) -> str:
-        if hasattr(obj.user, "profile") and obj.user.profile.discord_username:
-            return obj.user.profile.discord_username
-        return "-"
+    @admin.display(
+        description="Discord", ordering="user__profile__discord_member__username"
```
tim-schilling:
```suggestion
        description="Discord", ordering="user__profile__discord_member__nickname"
```

The nickname is what's shown in discord, so we should show that here instead.

home/admin.py
```diff
@@ -1024,6 +1027,38 @@ def has_add_permission(self, request) -> bool:
         return False
 
 
+@admin.register(DiscordMember)
+class DiscordMemberAdmin(admin.ModelAdmin):
+    """Read-mostly view of guild members mirrored from Discord."""
```
tim-schilling: Read-mostly?

home/admin.py
```diff
@@ -1024,6 +1027,38 @@ def has_add_permission(self, request) -> bool:
...
+
+    list_display = (
+        "username",
+        "nickname",
```
tim-schilling:
```suggestion
        "nickname",
        "username",
```
Let's lead with nickname since that's what people see in Discord.

home/forms.py
```diff
@@ -1291,39 +1291,32 @@ class Meta:
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         if self.instance.pk:
-            self.fields["discord_username"].initial = (
-                self.instance.user.profile.discord_username
+            self.fields["discord_member"].initial = (
+                self.instance.user.profile.discord_member_id
```
tim-schilling:
```suggestion
                self.instance.user.profile.discord_member
```

Shouldn't we just include the instance so it can render out the label properly?

home/models/discord.py
```diff
@@ -42,3 +41,29 @@ def __str__(self) -> str:
...
+
+    class Meta:
+        ordering = ["username"]
+
+    def __str__(self) -> str:
+        label = self.nickname or self.global_name or self.username
```
tim-schilling:
```suggestion
        label = self.nickname or self.username
```

I don't think there's a need show the global_name.

home/models/discord.py
```diff
@@ -42,3 +41,29 @@ def __str__(self) -> str:
...
+        default=True,
+        help_text="False when the member left the server; kept so profile "
+        "links survive leave/rejoin.",
+    )
+    last_seen_at = models.DateTimeField(default=timezone.now)
+
+    class Meta:
+        ordering = ["username"]
+
+    def __str__(self) -> str:
```
tim-schilling: Do you want to try using a [`GeneratedField`](https://docs.djangoproject.com/en/6.0/ref/models/fields/#generatedfield) for a `display_name` for this model? If we did that, it would remove the need to show both `nickname` and `username` everywhere. We could just reference `display_name` and be done with it easy peasy.

tim-schilling (commented):
Thanks @JohananOppongAmoateng for working on this so quickly! I've got lots of comments. Some of them are around the approach, so skim through them first and see which are discussions and which are things you can pass to an AI agent to have it fix.

The biggest one is around removing the linking flow. The other one is preferring `nickname` over `username` when possible.

tim-schilling:
@JohananOppongAmoateng I ended up having claude make an attempt to resolve the differences so we can get this sync shipped sooner.

tim-schilling:
I moved the discord models to the accounts app. I want to extend generating the User admin to include the changes from #870 once I merge that too.

tim-schilling:
Alright, I think this is good to be squashed and merged now.

tim-schilling:
Thank you @JohananOppongAmoateng! This will be really nice to have.

PR #869, opened 2026-08-14
Pass the user's IP address to buttondown when subscribing them.

### LINKED ISSUE #867: Pass IP address to Buttondown when subscribing an account
> In the last day, Djangonaut Space Transmissions added 23 subscribers through the API without including each subscriber's IP address.
> 
> When subscribers are created through the API from a server or serverless function — for example, a Supabase, Vercel, or Lambda integration — every request reaches us from your integration's IP address instead of the subscriber's. Our spam protection reads that as many signups from a single source and can start blocking legitimate subscribers.
> 
> To fix this, send the subscriber's real IP address in the `ip_address` field when you call `POST /v1/subscribers`. We'll use it to evaluate each subscriber on their own IP rather than your integration's shared one. Here's the reference: [Creating a subscriber](https://docs.buttondown.com/api-subscribers-create).
> 
> If you're not sure how to capture the subscriber's IP address, reply to this email and we'll help.

tim-schilling:
Do not store the user's IP address.

Fixes #867

### DIFF
--- a/accounts/receivers.py
+++ b/accounts/receivers.py
@@ -5,6 +5,7 @@ and for syncing users to the Buttondown newsletter.
 
 from home import constants
 from django.contrib.auth.models import Group
+from django.db import transaction
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 
@@ -63,7 +64,9 @@ def sync_buttondown_on_profile_save(
     if not buttondown_enabled() or raw or not instance.email_confirmed:
         return
     ip_address = getattr(instance, "_buttondown_ip_address", None)
-    sync_user_to_buttondown.enqueue(instance.user_id, ip_address=ip_address)
+    transaction.on_commit(
+        lambda: sync_user_to_buttondown.enqueue(instance.user_id, ip_address=ip_address)
+    )
 
 
 @receiver(
--- a/accounts/receivers.py
+++ b/accounts/receivers.py
@@ -55,10 +55,15 @@ def sync_buttondown_on_profile_save(
 
     Fires on every save where email_confirmed is True, keeping tags and
     subscription status in sync with any profile changes.
+
+    If the view that triggered this save set `_buttondown_ip_address` on the
+    instance (e.g. account activation), it's forwarded to the sync task so a
+    newly-created Buttondown subscriber records the originating request's IP.
     """
     if not buttondown_enabled() or raw or not instance.email_confirmed:
         return
-    sync_user_to_buttondown.enqueue(instance.user_id)
+    ip_address = getattr(instance, "_buttondown_ip_address", None)
+    sync_user_to_buttondown.enqueue(instance.user_id, ip_address=ip_address)
 
 
 @receiver(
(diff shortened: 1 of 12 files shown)


accounts/tests/test_activate_view.py
```diff
@@ -1,12 +1,19 @@
...
 from django.test import TestCase
+from django.test import override_settings
 from django.urls import reverse
 from django.utils.encoding import force_bytes
 from django.utils.http import urlsafe_base64_encode
 
 from accounts.factories import UserFactory
 from accounts.tokens import account_activation_token
 
+BD_SETTINGS = {"BUTTONDOWN_API_KEY": "test-api-key"}
```
tim-schilling: This should be patched in conftest.py

accounts/receivers.py
```diff
@@ -55,10 +55,15 @@ def sync_buttondown_on_profile_save(
...
+
+    If the view that triggered this save set `_buttondown_ip_address` on the
+    instance (e.g. account activation), it's forwarded to the sync task so a
+    newly-created Buttondown subscriber records the originating request's IP.
     """
     if not buttondown_enabled() or raw or not instance.email_confirmed:
         return
-    sync_user_to_buttondown.enqueue(instance.user_id)
+    ip_address = getattr(instance, "_buttondown_ip_address", None)
+    sync_user_to_buttondown.enqueue(instance.user_id, ip_address=ip_address)
```
tim-schilling: Wrap this in a `transaction.on_commit`?

docs/integrations/buttondown.md
```diff
@@ -57,3 +57,9 @@ uv run python manage.py sync_buttondown
 ```
 
 Ongoing syncs are handled automatically on every user profile save.
+
+## IP Address on Subscription
+
+Buttondown's subscriber-creation endpoint accepts an `ip_address` field, used for location detection and legitimacy validation. Since most syncs are triggered by `post_save` signals with no HTTP request in scope, the IP is only captured where a request is actually available: `ActivateAccountView` (`accounts/views.py`), the point at which most users first get synced to Buttondown.
+
+The flow: the view reads the client IP via `get_client_ip()` (`home/utils.py`) and stashes it as a transient, non-persisted attribute (`_buttondown_ip_address`) on the `UserProfile` instance before saving. The `post_save` signal handler (`accounts/receivers.py`) reads that attribute off the instance and forwards it through `sync_user_to_buttondown.enqueue()` -> `ButtondownService.sync_user()` -> `ButtondownClient.create_subscriber()`. It's only ever sent when a *new* Buttondown subscriber is being created — existing subscribers are updated via PATCH, which doesn't use this field. The IP is never persisted to our own database.
```
tim-schilling:
```suggestion
Buttondown's subscriber-creation endpoint accepts an `ip_address` field, used for location detection and legitimacy validation. When a user confirms their account (activates their account), the view stashes the IP address as a transient, non-persisted attribute (`_buttondown_ip_address`) on the `UserProfile` instance before saving. The `post_save` signal handle queues a task to sync the buttondown subscription.
```

PR #875, opened 2026-08-17
Speed up the test suite and CI execution

MarcusL11:
## Summary

This PR improves only the test-suite performance and CI execution.

### CI

The normal non-Playwright test suite improved from:

| | Before | After |
|---|---:|---:|
| Runtime | 95.21 seconds | 30.68 seconds |
| Tests | 1,110 | 1,107 |

This reduces the CI pytest runtime by **64.53 seconds**.

Across the normal tests, database-template build, and Playwright tests, the test related workflow steps improved from **4m 06s to 2m 45s**.

### Local profiling

The local default non-Playwright test profile improved from:

| | Before | After |
|---|---:|---:|
| Runtime | 44.90 seconds | 17.29 seconds |
| Tests | 1,090 | 1,090 |

_(New tests were added since this benchmark)_

This reduces the local runtime by **27.61 seconds**.

The largest measured improvement came from configuring `pytest-xdist` to use `loadscope`, which reduced the full-suite runtime from **44.90 seconds to 21.70 seconds**.

CI and local timings differ because CI uses four workers, collects coverage, and starts from a fresh environment. These are single-run measurements rather than statistically averaged results.

## Changes

- Added a dedicated test settings module with a faster password hasher, an immediate task backend, and disabled external service credentials.
- Reused the prebuilt test database template across regular and Playwright CI runs.
- Configured `pytest-xdist` to use the `loadscope` distribution strategy.
- Replaced `freezegun` with the faster `time-machine` library.
- Moved database-free slot tests to `SimpleTestCase`.
- Reused shared queryset fixtures with `setUpTestData`.
- Reduced repeated management command execution.
- Constrained pytest collection paths and enforced the Playwright marker boundary.
- Added a `just test-profile` command for identifying slow tests.


.github/workflows/tests.yml
```diff
@@ -71,18 +71,20 @@ jobs:
...
-        run: uv run pytest -n auto --cov --cov-branch --cov-report=xml --cov-report=term
-
       - name: Build test database template
         env:
           TEST_DB_TEMPLATE: template_wagtail_indymeet_test
         run: uv run python manage.py build_test_db_template
 
+      - name: Run tests
+        env:
+          TEST_DB_TEMPLATE: template_wagtail_indymeet_test
```
tim-schilling, line 81: I specifically chose to not use the template database for unit testing because it's more likely that a person would run into a database change error and encounter a weird error that's hard to diagnose

.github/workflows/tests.yml
```diff
@@ -71,18 +71,20 @@ jobs:
...
+        run: uv run pytest -n auto --dist loadscope --cov --cov-branch --cov-report=xml --cov-report=term
```
tim-schilling, line 82: TIL about loadscope. That's pretty neat

home/tests/test_queryset_managers.py
```diff
@@ -100,30 +100,28 @@ def test_for_user_private_event_without_session_not_visible_to_non_member(self):
 class UserSurveyResponseQuerySetTestCase(TestCase):
     """Test UserSurveyResponseQuerySet methods."""
 
-    def setUp(self):
-        """Create test data using factories."""
-        self.session = SessionFactory(
+    @classmethod
+    def setUpTestData(cls):
```
tim-schilling, line 104: My understanding of `setUpTestData` when I first started writing these tests was incomplete. I think we should be using `setUpTestData` way more often than we are. Pretty much anything that we use `setUp` for we should be able to move up into `setUpTestData`. In short: this is a good change, but I think we should have more.

pyproject.toml
```diff
@@ -67,6 +67,21 @@ dev = [
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "indymeet.test_settings"
 addopts = "--reuse-db -m 'not playwright'"
+testpaths = [
+    "accounts",
+    "home",
+    "tests",
+]
```
tim-schilling, line 74: Can you explain to me the purpose of this one? My worry is that this is another thing we have to remember if we add a new app. This would be helpful context to include in the commit message.

pyproject.toml
```diff
@@ -66,7 +66,7 @@ dev = [
 
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "indymeet.test_settings"
-addopts = "--reuse-db -m 'not playwright'"
+addopts = "--reuse-db -m 'not playwright' --strict-markers"
```
tim-schilling, line 69: Similar to another comment, I'm not sure I understand the "why" on this one, due to my lack of knowledge. The context may be helpful to include on the commit message or a comment.

accounts/management/commands/tests/test_notify_unusable_password_users.py
```diff
@@ -16,55 +14,26 @@ def settings(self, settings):
...
-        user.set_unusable_password()
-        user.save()
-
-        management.call_command("notify_unusable_password_users")
-
-        assert len(mail.outbox) == 1
-        assert "/accounts/reset/" in mail.outbox[0].body
-
-    def test_email_contains_delete_account_url(self):
+    def test_sends_expected_email_to_unusable_password_user(self):
```
tim-schilling, line 17: Oh, thank you. This type of AI test bloat bothers me, but I rarely catch it.

tim-schilling (commented):
This looks pretty good! I'm a little worried about fully committing to the test template db, but I think I'd be fine experimenting with it.

.github/workflows/tests.yml
tim-schilling: Ignore this. I thought this was in the justfile

pyproject.toml
MarcusL11: Sorry! This was added while measuring test speed to limit pytest collection to the current test locations, and I forgot to remove it afterward.

pyproject.toml
MarcusL11: --strict-markers makes pytest fail when a marker is misspelled or unregistered, instead of continuing with only a warning. It does not improve performance. 

The project currently has only one custom marker and pytest already emits an unknown marker warning, so I can see the argument that this is not needed. 

Let me know, happy to remove it :)

accounts/management/commands/tests/test_notify_unusable_password_users.py
MarcusL11: ai catching ai 

image

pyproject.toml
tim-schilling: Nope, that makes sense. Thank you for the explanation

tim-schilling (approved):

tim-schilling:
@MarcusL11 looks like there's a error in the tests.

MarcusL11:
@tim-schilling  fixed, it was leftovers freezgun module !

tim-schilling (approved):
Alright I'm good with this. I'd like to try to keep this as separate commits, but I want to consolidate that merge commit out. Do you have any interest in cleaning this up and rebasing on main? I'm happy to do that otherwise. It's a peculiarity of mine.

MarcusL11:
@tim-schilling yes, absolutely-- will put that fix in.

tim-schilling:
Thank you @MarcusL11!

PR #876, opened 2026-08-18
Improve the timezone selection UX

### LINKED ISSUE #865: Improve the timezone selection UX
The current timezone selector is pretty rough and includes timezones we probably shouldn't such as the US/* values. Ideally we'd have these timezones categorized by the hour offset and/or have a autocomplete search utility.

Image

JohananOppongAmoateng:
Closes #865


tim-schilling:
Thanks for creating this @JohananOppongAmoateng! With PRs like this one where it changes the look of the application, the PR should include screenshots of what the difference is. This allows for a quicker review of the visuals before getting into the code.

tim-schilling:
@JohananOppongAmoateng can you explain the approach you're suggesting we take and if there's a reason why you prefer this too? I'd like to understand more of your thinking if that's alright.

JohananOppongAmoateng:
@tim-schilling  okay so i am going to attach screenshots and how i understood the problem soon

JohananOppongAmoateng:
Screenshot_20260824_194507
The old dropdown was just every timezone name dumped in a list. It had stuff we don’t need like US/Eastern.
I made it a list of UTC offsets instead. So you pick UTC-05:00, not America/New_York.
For availability we care about the offset, not the city. Grouping cities under offsets still leaves a huge list. Search felt like extra if we’re not showing cities.
US/* goes away because we’re not listing zones at all.

tim-schilling:
@JohananOppongAmoateng gotcha, thank you for the explanation. Unfortunately, I don't think we should expect people to know their UTC offset. It's not a very common thing unless you're scheduling meetings internationally. Additionally, the utc offset ignores timezones that may or may not utilize daylight savings time. We need to capture the city for their timezone which leaves us back at supporting some type of search for the drop down.

JohananOppongAmoateng:
@tim-schilling  so if i undertsand what you mean you want to maintain the old version and allowing users to be able to search right or have some searchanle dropdown?

tim-schilling:
Yes, I think you understand my concerns correctly. I'm open to other implementations, it doesn't need to be a dropdown if you have a different idea in mind.

PR #881, opened 2026-08-27
Updated info in the homepage

raffaellasuardini:
As Dan suggested, the homepage information about the current "6th session" has been updated.


tim-schilling (approved):
@raffaellasuardini I think changes like this you are good to merge without a review

PR #886, opened 2026-09-04
Support evaluating tutorials automatically.

tim-schilling:


home/models/tutorial_evaluation.py
tim-schilling, line 1: Rename this to remove github_ prefix.

home/models/github_tutorial_evaluation.py
```diff
@@ -0,0 +1,81 @@
+"""Scheduling and results for automated make_toast tutorial evaluations.
+
+Applicants link their completed contribution tutorial submission as part of
+their session application. Evaluating a link hits the GitHub API (see
+``home.integrations.github.evaluate_submission``), so it isn't done inline on
+every save. Instead, ``home.tasks.github_tutorial_evaluations`` queues a
+re-check for a day out whenever a survey response is saved with a non-passing
+result, and a daily cron job works whatever is due.
```
tim-schilling:
```suggestion
Applicants link their completed contribution tutorial submission as part of
their session application. Evaluating a link hits the GitHub API (see
``home.integrations.github.evaluate_submission`` and ``home.tasks.github_tutorial_evaluations``).
```

home/tasks/tutorial_evaluations.py
tim-schilling, line 1: Rename accordingly to the other change.

home/tasks/github_tutorial_evaluations.py
```diff
@@ -0,0 +1,170 @@
...
+there's still time left in the application window. It's a plain function,
+not something hooked into ``UserSurveyResponse.save()``: callers (currently
+the create/edit survey response views) invoke it explicitly alongside the
+save, the same way those views already trigger the application notification
+emails. A daily cron job then works whatever has come due, re-evaluating the
+linked submission and, if it's still not a pass, sending a one-time reminder
+email. Calling ``schedule_evaluation_if_needed`` again later (e.g. after the
+applicant fixes their link) queues another check the same way.
+"""
+
```
tim-schilling:
```suggestion
```

home/tasks/tutorial_evaluations.py
```diff
@@ -0,0 +1,170 @@
...
+from django.utils import timezone
+from django_tasks import task
+
+from home import email
+from home.integrations.github import evaluate_submission
+from home.models import Session, TutorialEvaluation, UserSurveyResponse
+from home.models.github_tutorial_evaluation import Result
+
+# The key of the application survey question that collects the tutorial
+# submission link, derived from its label via Question.save()'s slugify().
```
tim-schilling, line 15:
```suggestion
# submission link, derived from its label via Question.save()'s slugify().
# HACK: This is intentional and fragile, sorry
```

home/tasks/github_tutorial_evaluations.py
```diff
@@ -0,0 +1,170 @@
...
+        return
+
+    evaluation, _ = TutorialEvaluation.objects.get_or_create(
+        user_survey_response=user_survey_response
+    )
+    if evaluation.result == Result.CORRECT:
+        return
+
+    evaluation.next_evaluation_at = next_check
+    evaluation.save(update_fields=["next_evaluation_at", "updated_at"])
```
tim-schilling: This should be in an atomic transaction block to avoid race conditions with a `reevaluate_tutorial_submission` task.

home/tasks/github_tutorial_evaluations.py
```diff
@@ -0,0 +1,170 @@
...
+from home.models import Session, TutorialEvaluation, UserSurveyResponse
+from home.models.github_tutorial_evaluation import Result
+
+# The key of the application survey question that collects the tutorial
+# submission link, derived from its label via Question.save()'s slugify().
+TUTORIAL_QUESTION_KEY = (
+    "please-complete-the-django-contributing-tutorial-and-provide-a-link-below"
+)
+
+RE_EVALUATION_DELAY = datetime.timedelta(days=1)
```
tim-schilling:
```suggestion
REEVALUATION_DELAY = datetime.timedelta(days=1)
```

home/tasks/github_tutorial_evaluations.py
```diff
@@ -0,0 +1,170 @@
...
+    evaluation.reasons = list(result.reasons)
+    evaluation.notes = result.notes
+    evaluation.evaluated_at = timezone.now()
+    evaluation.next_evaluation_at = None
+    evaluation.save()
+
+    if evaluation.result != Result.CORRECT:
+        _send_reminder(survey_response)
+        evaluation.reminder_sent_at = timezone.now()
+        evaluation.save(update_fields=["reminder_sent_at", "updated_at"])
```
tim-schilling: Should this be extracted out? If this fails it will roll back the whole transaction rather than just part of it, meaning the evaluation won't have been saved and the `reminder_sent_at` wasn't set.

tim-schilling: I think this data flow is a bit wonky. Perhaps it should work as follows.

When a survey response is submitted, the evaluation task is scheduled immediately. Then we have a task that looks at the `.due()` and those that aren't correct along with `reminder_sent_at` to determine which reminders to send. That way we separate the flow into two queues. The first is just about evaluating responses, both immediately after and another to check daily in case there were errors. The other queue just processes the emails that need to be sent. I think that simplifies the application flow and makes things more robust.

home/views/team_formation.py
```diff
@@ -268,6 +268,14 @@ def get_filtered_applicants(
         if prev_avg_score is not None:
             prev_avg_score = round(prev_avg_score, 1)
 
+        # Get the tutorial evaluation result, if the submission has been evaluated
+        evaluation = getattr(response, "github_tutorial_evaluation", None)
+        tutorial_result = (
+            evaluation.get_result_display()
```
tim-schilling, line 274: Where is this defined?

home/views/team_formation.py
```diff
@@ -268,6 +268,14 @@ def get_filtered_applicants(
         if prev_avg_score is not None:
             prev_avg_score = round(prev_avg_score, 1)
 
+        # Get the tutorial evaluation result, if the submission has been evaluated
+        evaluation = getattr(response, "github_tutorial_evaluation", None)
```
tim-schilling: Do we need a select_related for this? If possible, can we use a `.annotate()`?

home/forms.py
```diff
@@ -446,7 +446,8 @@ def generate_full_csv(self, survey: Survey) -> HttpResponse:
         header.extend([q.label for q in questions])
         # Add scorer columns
         header.extend([f"{name} score" for name in scorer_names])
-        # Add Score and Selection Rank columns
+        # Add Tutorial Result, Score, and Selection Rank columns
+        header.append("Tutorial Result")
```
tim-schilling: Move the tutorial result in front of the scores.

PR #900, opened 2026-09-14
Speed up team auto-allocation and allocate by rank tier.

tim-schilling:
The auto-allocate admin action never finished on real sessions. The search tried every candidate/team assignment without pruning, copied the full state at every step, and recomputed availability overlaps from Slot objects each time.

- Prune branches that can't beat the best allocation found so far. Equal scores never replaced the best state, so results are unchanged.
- Update one working state in place and undo each move, copying only when a new best allocation is found.
- Compare availability as UTC floats computed once per user, and update each team's shared meeting window as djangonauts are added/removed.
- Check project preference and captain overlap once per candidate/team pair instead of at every step.
- Fetch candidates once, loading their availability in the same query.
- Allocate ranks 0-1 first, then rank 2, then rank 3, each only while teams have open slots.
- Add AllocationState.placements_by_rank, and show the placed/considered counts per rank in the admin success message.
- unallocated_candidates now lists exactly the eligible applicants who weren't placed.


tim-schilling:
I'm going to merge this as it's working locally with a clone of the session and we need to unblock the organizing.

